PR #8, opened 2018-03-07
Fix create user local cr3.ini

slonm:
On the Linux OS after install app cr3.ini does not exist. App trying to read user local ini (/home/<user>/.cr3/cr3.ini) then system level ini (/usr/share/cr3/cr3.ini). The last one cached and on save settings used cached value. When system level ini is not writable the app lost settings at the exit.
Fix store in the cache only writable ini name

### DIFF
--- a/cr3qt/src/mainwindow.cpp
+++ b/cr3qt/src/mainwindow.cpp
@@ -100,8 +100,13 @@ MainWindow::MainWindow(QWidget *parent)
 
     ldomDocCache::init( qt2cr( cacheDir ), DOC_CACHE_SIZE );
     ui->view->setPropsChangeCallback( this );
+    bool settingsExists = false;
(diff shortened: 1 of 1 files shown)


cr3qt/src/mainwindow.cpp
```diff
@@ -100,8 +100,13 @@ MainWindow::MainWindow(QWidget *parent)
 
     ldomDocCache::init( qt2cr( cacheDir ), DOC_CACHE_SIZE );
     ui->view->setPropsChangeCallback( this );
+		bool settingsExists = false;
     if ( !ui->view->loadSettings( iniFile ) )
-        ui->view->loadSettings( iniFile2 );
+        settingsExists = ui->view->loadSettings( iniFile2 );
```
Frenzie: I think @buggins wants spaces, not tabs.

PR #29, opened 2018-08-07
Fixed invalid text measurements when using HarfBuzz and in used text …

virxkane:
Fix for my own kerning implementation, sorry for broken code...
Fixed invalid text measurements when using HarfBuzz and in used text font found any ligatures (for example, "Noto Serif").
The key error: fill char's flags at invalid position:

-flags[i] = GET_CHAR_FLAGS(ch);
+flags[cluster] = GET_CHAR_FLAGS(ch);

Invalid:
screenshot_20180807_080008
Fixed:
screenshot_20180807_102410

### DIFF
--- a/crengine/src/lvfntman.cpp
+++ b/crengine/src/lvfntman.cpp
@@ -1244,11 +1244,13 @@ public:
             glyph_count = hb_buffer_get_length(_hb_buffer);
             glyph_info = hb_buffer_get_glyph_infos(_hb_buffer, 0);
             glyph_pos = hb_buffer_get_glyph_positions(_hb_buffer, 0);
+#ifdef _DEBUG
             if (glyph_count != len) {
-                CRLog::info(
+                CRLog::debug(
                         "measureText(): glyph_count not equal source text length (ligature detected?), glyph_count=%d, len=%d",
                         glyph_count, len);
             }
+#endif
             register int j;
             register uint32_t cluster;
             register uint32_t prev_cluster = 0;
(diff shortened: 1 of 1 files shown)


poire-z:
Just reporting some related additional fixes for stuff like this, as I got because of ligatures:
<kbd>image</kbd>

Got that on koreader, and we do additional stuff in lvtextfm.cpp, so dunno if you would get that too with coolreader, but I think you would.
It happened because of bad flags at end of measureText segment, because the bold node make out another segment to measure. Bad flags at the end of the first segment were causing this.

No real patch, just some extracts from mine, around that code in this PR:

```diff
             uint32_t j;
             uint32_t cluster;
             uint32_t prev_cluster = 0;
+            int skipped_chars = 0; // to add to 'i' at end of loop, as 'i' is used later and should be accurate
             for (i = 0; i < (int)glyph_count; i++) {
```

```diff
                 for (j = prev_cluster + 1; j < cluster; j++) {
                     // fill flags and widths for chars skipped (because they are part of a
                     // ligature and are accounted in the previous cluster)
                    flags[j] = GET_CHAR_FLAGS(text[j]);
                    widths[j] = prev_width;            // for chars replaced by ligature
+                    skipped_chars++;
                 }
```

```diff
                 for (j = prev_cluster + 1; j < (uint32_t)len; j++) {
                     flags[j] = GET_CHAR_FLAGS(text[j]);
                     widths[j] = prev_width;
+                    skipped_chars++;
                 }
             }
+            // i is used below to "fill props for rest of chars", so make it accurate
+            i += skipped_chars;
```

```diff
         // fill props for rest of chars
         for ( int ii=i; ii<len; ii++ ) {
-            flags[i] = GET_CHAR_FLAGS( text[ii] );
+            flags[ii] = GET_CHAR_FLAGS( text[ii] );
         }
```

Hope it will save you the many hours it took me to find out that last typo :) (guess that code didn't get triggered before, but is now with harfbuzz as, with skipped chars, there are now _rest of chars_ to fill).

virxkane:
@poire-z Ok, thank you, I fix this later.

PR #35, opened 2018-08-14
Fixed regression in commit ced047b1048dc31c006d7262ea4f8cd29f66f805 a…

virxkane:
Fixed regression in commit ced047b1048dc31c006d7262ea4f8cd29f66f805 at 2018-08-09 (Fixed stack corruption on broken files: index out of range in PreProcessXmlString() in lvxml.cpp)

Sequences like '&nbsp;', '&amp;' and other not performed as should.


virxkane:
Sequences like '& nbsp ;' ; '& amp ;' and other not performed as should.

PR #70, opened 2019-01-24
Not loading ePub files

ThaneGreyhaven:
God damn.  I think I finally found a way to send you messages about the VR Cool Reader app for Oculus Go.

This wasn't easy to find.  If you want feedback, you've got to make this easier.  I love the app, I really do.  It is by far the best eReader for the Go, hands down, I especially like the page turning animation, that is very good to see.  It does have a few problems though, and I've got some suggestions.

1) I've loaded all my ePubs onto the Go (just under 2600 of them) and it won't find them.  Is there somewhere specific I should put them?  Because  if there is, that should be indicated somewhere.  I've tried putting them in Documents and Downloads and it hasn't picked them up.  Maybe have a 'Browse for Library' feature.

2) The movement inside of the library is VERY clunky, there is a touchpad on the controller and swiping left or right should turn you around, rather than the arrows on the floor.

3) 'Drop the book' should be changed to 'Reshelve the book' and it should immediately go back on the shelf.  Concurrently you should get rid of 'Close the book' as that has no useful purpose (even if it is kind of neat). The problem is that once you close the book, you can't do anything with it except open it again as the book blocks all other buttons.

4) Not sure what the 'Retrieve' function is supposed to accomplish, but I think that should be changed to 'Resize'.  There should be controls to change the size of the book and the size of the font in the book.  Most eReaders allow this and some even allow for changing the spacing of the letters (aka kearning) and how much space the margins take up, how much extra space is at first line of a paragraph, the font being used, the font colour, the page colour etc etc etc. 

5) Last and most minor suggestion.  It would be good to have different views out of the windows.

I would imagine that these (and many other) ideas are already being thought of and worked on, but I wanted to make sure that you were aware of them in case they hadn't occurred to you.  Please do keep them in mind and keep up the good work on this!


ThaneGreyhaven:
Also wanted to mention that I think you should have an option to turn off the worms.  I love the concept, but TONS of people have phobias about them, so an ability to turn them off would be a good idea.

buggins:
Closing as outdated, a lot of merge conflicts.

PR #85, opened 2019-03-18
Hb light features

virxkane:
Fixed bug with text rendering without ligatures on some fonts. Thanks to koreader project members:
https://github.com/koreader/koreader/pull/4744
https://github.com/koreader/crengine/pull/271
The essence of the bug: some characters, such as the “è” symbol, are divided into two glyphs, which are prohibited when ligatures are turned off.
before

Affected font - 'FreeSerif.ttf' in https://github.com/koreader/koreader-fonts/tree/master/freefont
Test file: [Da.Vinci.Code.FR.epub.zip](https://github.com/buggins/coolreader/files/2976730/Da.Vinci.Code.FR.epub.zip)

### DIFF
--- a/cr3qt/src/settings.cpp
+++ b/cr3qt/src/settings.cpp
@@ -139,7 +139,11 @@ SettingsDlg::SettingsDlg(QWidget *parent, CR3View * docView ) :
     m_ui->cbEnableEmbeddedFonts->setEnabled(m_props->getBoolDef(PROP_EMBEDDED_STYLES, true));
     optionToUi( PROP_TXT_OPTION_PREFORMATTED, m_ui->cbTxtPreFormatted );
     optionToUi( PROP_FONT_KERNING_ENABLED, m_ui->cbFontKerning );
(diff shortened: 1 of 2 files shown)


poire-z:
About the `lastFitChar = i + 1;`, it could also help in the non-harfbuzz code (I don't know if Coolreader still uses it), after your line 1611:
```diff
            if ( w==0xFF ) {
                glyph_info_t glyph;
                if ( getGlyphInfo( ch, &glyph, def_char ) ) {
                    w = glyph.width;
                    _wcache.put(ch, w);
                } else {
                    widths[i] = prev_width;
+                   lastFitChar = i + 1;
                    continue;  /* ignore errors */
}
```
(We fixed that in https://github.com/koreader/crengine/commit/82624eedb6e92f814a837f83757916f72fafa794 before your triplet code - I just reported that fix when integrating your code, and forgot to tell you at the time.)

virxkane:
@poire-z thank you, fixed.

PR #95, opened 2019-05-13
Build and upload AppImage

probonopd:
This PR, when merged, will compile this application on [Travis CI](https://travis-ci.org/) upon each `git push`, and upload an [AppImage](http://appimage.org/) to your GitHub Releases page.

Providing an [AppImage](http://appimage.org/) would have, among others, these advantages:
- Applications packaged as an AppImage can run on many distributions (including Ubuntu, Fedora, openSUSE, CentOS, elementaryOS, Linux Mint, and others)
- One app = one file = super simple for users: just download one AppImage file, [make it executable](http://discourse.appimage.org/t/how-to-make-an-appimage-executable/80), and run
- No unpacking or installation necessary
- No root needed
- No system libraries changed
- Works out of the box, no installation of runtimes needed
- Optional desktop integration with `appimaged`
- Optional binary delta updates, e.g., for continuous builds (only download the binary diff) using AppImageUpdate
- Can optionally GPG2-sign your AppImages (inside the file)
- Works on Live ISOs
- Can use the same AppImages when dual-booting multiple distributions
- Can be listed in the [AppImageHub](https://appimage.github.io/) central directory of available AppImages
- Can double as a self-extracting compressed archive with the `--appimage-extract` parameter
- No repositories needed. Suitable/optimized for air-gapped (offline) machines
- Decentralized

[Here is an overview](https://appimage.github.io/apps) of projects that are already distributing upstream-provided, official AppImages.

__PLEASE NOTE:__ For this to work, you need to set up `GITHUB_TOKEN` in Travis CI for this to work; please see https://github.com/probonopd/uploadtool.

If you have questions, AppImage developers are on #AppImage on irc.freenode.net.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,32 @@
+language: cpp
+compiler: gcc
+sudo: require
+dist: trusty
+
+before_install:
+  - sudo add-apt-repository ppa:beineri/opt-qt593-trusty -y
+  - sudo apt-get update -qq
+
+install:
+  - sudo apt-get -y install pkg-config qt59base qt59tools libpng12-dev libfreetype6-dev libjpeg62-dev libfontconfig1-dev zlib1g-dev
+  - source /opt/qt*/bin/qt*-env.sh
+
+script:
+  - sudo ln -s /usr/include/freetype2/ft2build.h /usr/include/ # https://github.com/matplotlib/matplotlib/issues/3029#issuecomment-43318941
+  - mkdir qtbuild
+  - cd qtbuild
+  - cmake -DGUI=QT5 -DCMAKE_CXX_FLAGS="-DUSE_FREETYPE2" -DCMAKE_BUILD_TYPE=Release -DMAX_IMAGE_SCALE_MUL=2 -DDOC_DATA_COMPRESSION_LEVEL=3 -DDOC_BUFFER_SIZE=0x1400000 -DCMAKE_INSTALL_PREFIX=/usr ..
+  - make -j$(nproc)
+  - make DESTDIR=appdir -j$(nproc) install ; find appdir/
+  - wget -c -nv "https://github.com/probonopd/linuxdeployqt/releases/download/continuous/linuxdeployqt-continuous-x86_64.AppImage"
+  - chmod a+x linuxdeployqt-continuous-x86_64.AppImage
+  - ./linuxdeployqt-continuous-x86_64.AppImage appdir/usr/share/applications/*.desktop -appimage
+
+after_success:
+  - wget -c https://github.com/probonopd/uploadtool/raw/master/upload.sh
+  - bash upload.sh Cool*.AppImage*
+  
+branches:
+  except:
+    - # Do not build tags that we create when we upload to GitHub Releases
+    - /^(?i:continuous)/


probonopd:
One thing that is not correct seems to be

```
2019/05/13 19:43:15.9970 ERROR Canot load translation file cr3_en_US from dir /usr/share/cr3/i18n/
2019/05/13 19:43:16.0825 ERROR Document cache index file cannot be read
2019/05/13 19:43:16.0972 ERROR Error opening file /usr/share/cr3/cr3.ini for reading
2019/05/13 19:43:16.0973 ERROR Cannot load settings from file /usr/share/cr3/cr3.ini
2019/05/13 19:43:16.0982 ERROR Error opening file /usr/share/cr3/cr3hist.bmk for reading
2019/05/13 19:43:16.0982 ERROR Error opening file /usr/share/cr3/fb2.css for reading
```

Instead of loading stuff from a hardcoded fixed loction at `/usr/share/`, these paths should be constructed relative to the main executable directory, e.g., using `QCoreApplication::applicationDirPath()`. Do you think this could be done?

Then it would work also when executed from locations such as `/tmp/.mount_CoolRe3iOBNY/usr/bin/coolreader`, and would load its resources from `/tmp/.mount_CoolRe3iOBNY/usr/share/coolreader/...`.

http://doc.qt.io/qt-5/qcoreapplication.html#applicationDirPath

virxkane:
@probonopd 

> Instead of loading stuff from a hardcoded fixed loction at /usr/share/, these paths should be constructed relative to the main executable directory, e.g., using QCoreApplication::applicationDirPath(). Do you think this could be done?

I think this is not correct in common case, because FHS:  https://en.wikipedia.org/wiki/Filesystem_Hierarchy_Standard .
But you can patch sources and implementing loading resources relative to application executable directory, but only for AppImage builds: you can add cpp directive, for example `-DBUILD_APPIMAGE=1`, to CMAKE_CXX_FLAGS and check this in source code with `#if`, `#ifdef`.

probonopd:
Actually this would not violate the FHS, because if the prefix to which the application is installed happens to be `/usr`, then everything is like before.

buggins:
Guys, is it safe to merge this PR?

virxkane:
@buggins looks harmless enough, but require setting up Travis CI and other...

virxkane:
I test the resulting AppImage in Ubuntu 19.04 and it works.
https://github.com/probonopd/coolreader/releases/tag/continuous

probonopd:
Thanks for merging @buggins but you also need to enable Travis CI and set a token as described in https://github.com/probonopd/uploadtool#usage, otherwise the builds will not show up on https://github.com/buggins/coolreader/releases.

https://travis-ci.com/buggins/coolreader/ needs to be enabled.

exander77:
AppImages are not built. Is the Travis set up?

probonopd:
In the meantime, Travis CI seems to have changed for the worse. Probably GitHub Actions is now preferable.

exander77:
Btw, that 2019 version from your fork is working fine, I would build it myself, but having a kind of CI that builds official AppImaga sounds preferable to me.

PR #111, opened 2019-10-06
Font render fixes

virxkane:
1. Fixed regression: incorrect text rendering when using fallback font - wrong text measurements or even freezes. Thanks to konic2884 at https://4pda.ru/forum for log-file.
2. Refactoring: files lvfntman.h, lvfntman.cpp are divided into several, for each class a separate file, for more convenient work.
3. Desktop Qt5: reflect style changes in preview widget of setting dialog: kerning, ligatures, font hinting, font gamma, floating punctuation.
4. Improved glyph caching: instead of a linked list a hash table is used, as result significally faster item's lookup, but, unfortunately, this does not affect the overall performance ...
Test program with benchmarking in crengine/Tools/glyphcache_bench/.
This feature can be enabled/disabled in crengine/include/crsetup.h via symbol USE_GLYPHCACHE_HASHTABLE.


crengine/src/private/lvfreetypeface.cpp
```diff
@@ -1064,7 +1066,7 @@ LVFontGlyphCacheItem *LVFreeTypeFace::getGlyph(lUInt16 ch, lChar16 def_char) {
 
 LVFontGlyphCacheItem* LVFreeTypeFace::getGlyphByIndex(lUInt32 index) {
     //FONT_GUARD
-    LVFontGlyphCacheItem *item = _glyph_cache2.getByIndex(index);
+    LVFontGlyphCacheItem *item = _glyph_cache2.get(index);
```
poire-z: Is this rename voluntarily?
I see you renamed `getByChar` (that you renamed from `get` in 3rd commit) back to `get` - but you did not for `getByIndex`.
(Also, on linux lChar16 = wchar_t = 32 bits on Linux, just like lUInt32 = 32 bits - but may be not on Windows I think - so, somehow, even for glyph index, you could use just one method - may be that's what you aimed at?)

virxkane:
Just reverted back. New function getByIndex() just for better code readability.

poire-z:
Sorry, I wasn't clear.
Just greping the crengine code for "getBy" (or "get") and removing unrelated stuff, to get some nice symetrical pairs of stuff:

After your 3rd commit, you have (just grep'ing) in the whole code:
```
LVFontGlyphCacheItem *LVFontLocalGlyphCache::getByChar(lChar16 ch) {
LVFontGlyphCacheItem*LVFontLocalGlyphCache::getByIndex(lUInt32 index)
LVFontGlyphCacheItem *getByChar(lChar16 ch);
LVFontGlyphCacheItem *getByIndex(lUInt32 index);
LVFontGlyphCacheItem *item = _glyph_cache.getByChar(ch);
LVFontGlyphCacheItem *item = _glyph_cache2.getByIndex(index);
```

After this 5th commit, where you renamed `getByChar` back to `get`:
```
LVFontGlyphCacheItem *LVFontLocalGlyphCache::get(lChar16 ch) {
LVFontGlyphCacheItem*LVFontLocalGlyphCache::getByIndex(lUInt32 index)
LVFontGlyphCacheItem *get(lChar16 ch);
LVFontGlyphCacheItem *getByIndex(lUInt32 index);
LVFontGlyphCacheItem *item = _glyph_cache.get(ch);
LVFontGlyphCacheItem *item = _glyph_cache2.get(index);  <= this one feels like an error
```

The last one feels like you renamed too much, and it should stay `_glyph_cache2.getByIndex`.
(And it might just work because on Linux, lChar16 and lUInt32 are both the same 32 bits thingie - and so, this rename error might be unnoticed.)

virxkane:
@poire-z Yes of course, this is typo, thank you. I did not immediately understand what you mean, sorry.
> you could use just one method - may be that's what you aimed at?

Maybe, but I want to separate characters from glyph indices.

crengine/include/crsetup.h
```diff
@@ -300,6 +304,10 @@
 #define MAX_IMAGE_SCALE_MUL 2
 #endif
 
+#ifndef USE_GLYPHCACHE_HASHTABLE
+#define USE_HASHTABLE_GLYPHCACHE 0
```
poire-z: Typo in that #define.

virxkane: Thank you, fixed too.

PR #119, opened 2019-12-11
Fixed docx parser to handle properly a case when only content element in a paragraph is hyperlink

pkb:
The following example illustrates the problem:
```xml
	<w:p>
	               <w:pPr>
                         ...
			</w:pPr>
			<w:hyperlink w:anchor="bookmark238" w:tooltip="Current Document">
				<w:r>
					<w:rPr>
					</w:rPr>
					<w:t>Some item</w:t>
				</w:r>
				<w:r>
					<w:rPr>
					</w:rPr>
					<w:tab/>
					<w:t>27</w:t>
				</w:r>
			</w:hyperlink>
	</w:p>
```
Before the fix, no `<p>` element was produced for the above code, because of it, all such paragraphs were glued together on a single line and it is not good for reading :)

@poire-z, FYI this bug was reported by koreader user.

### DIFF
--- a/crengine/src/docxfmt.cpp
+++ b/crengine/src/docxfmt.cpp
@@ -1912,6 +1912,7 @@ ldomNode * docx_pHandler::handleTagOpen(int tagId)
 {
     switch(tagId) {
     case docx_el_r:
+    case docx_el_hyperlink:
         if ( 0 == m_runCount ) {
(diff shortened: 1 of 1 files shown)


poire-z:
Thanks for the ping and fix, I'll cherry-picked these commits in a few days.

PR #125, opened 2020-01-18
Merged with KOReader fork of crengine

### LINKED ISSUE #91: Merge with KOReader fork of crengine
KOReader developers (ping @Frenzie @poire-z) maintain a fork of crengine: https://github.com/koreader/crengine/ Quick diff shows that there are list item display fixes (see #90), several new CSS properties support, fix for `<cite>` bug (https://sourceforge.net/p/crengine/bugs/268/) and other enhancements. Are there any plans to merge these fixes back into coolreader?

Related issue about another fork: https://github.com/koreader/crengine/issues/267

(KOReader is great, but it focuses on e-ink devices, so it does not even support changing background color. For this reason, it can't be used as a drop-in replacement for CoolReader on Android.)

pkb:
This PR should close #91. It includes a lot of enhancements in css processing, thanks to koreader devs and especially to poire-z for work on crengine.

It wasn't that easy to merge, in case of conflicts I mostly accepted their side. Except for kerning mode setting, We handle it differently - we use a combination of new setting "Text shaping mode" and the old Kerning enable/disable. 

They have added qimagescale from QT, I moved it to thirdpatrty folder. I also added nanosvg library to the thirdparty folder, harfbuzz I updated to 2.6.4 version.

**Note:** this change is not yet ready for Android, @virxkane will create a separate request for it.


poire-z:
I see you haven't brought our epub.css, which might be fine, dunno.
But you should be aware of some other related changes I made that may cause a risk to your long time users : I made some changes to the default `display: block / inline` of some elements (in fb2def.h, and in our epub.css), and this might change the DOM tree - and so, xpath/xpointers that your users may have saved might not be found in the new DOM.
Some comments about that in: https://github.com/pkb/coolreader/blob/eccef73becff11af69a27dd763cf4991385b7cda/crengine/src/lvtinydom.cpp#L15-L65
And in https://github.com/koreader/crengine/pull/172 https://github.com/koreader/crengine/pull/192 https://github.com/koreader/crengine/pull/194 (and possibly others...)

On KOReader side, we save a `cre_dom_version` setting, and previous books opened with an older version have koreader request a specific DOM version handling from the code, in the hope old xpointers will be found in a DOM tree built with new tweaks/fixes reverted.
This also affected our epub.css, that has this ugly section at its bottom:
https://github.com/koreader/crengine/blob/f20454111fff249d78a1f092ff7ccba52c8b720f/cr3gui/data/epub.css#L170-L207

I remember that I have thought a few times "they will have to take care of that if it's brought back into upstream" - but right now, I can't remember the specific points, except the one above.
But I usually tried to keep upstream (your :) behaviour as the default (often mentionned as "legacy" in the code), and our enhancements are enabled from our frontend - via some of [the added PROP_*](include/lvdocviewprops.h) and some other possibly via specific added methods (but I can't find any in lvdocview.cpp :|) or via added parameters to existing methods, keeping the legacy behaviour with the default argument for the added parameter.
For example, float support (that may break xpath/xpointers) is enabled via `PROP_RENDER_BLOCK_RENDERING_FLAGS "crengine.render.block.rendering.flags"` - which should default to 0 (legacy rendering), but if you want to enable floats and the other block rendering flags, you'll have to add settings/toggles to your various frontends.

I can't remember about all the things I added over the last 2 years and which may affect you - but the PRs I opened in koreader/crengine, and my commit messages, are usually quite verbose :) so you might want to spend some time browsing thru them to see what you get with this merge (possible issues, or enhancements that need to be enabled from frontends).

Another (a lot more concise :) source for what's available for your frontends might be the history of our single-file crengine interface: https://github.com/koreader/koreader-base/commits/master/cre.cpp

TLDR: be careful, and test a lot before making a release with this :)

Frenzie:
Having the two come closer together is a welcome and doubtlessly quite time consuming endeavor. :-)

pkb:
> But you should be aware of some other related changes I made that may cause a risk to your long time users : 

@poire-z thank you very much for bringing this to our attention.

poire-z:
Sharing more thoughs (as they are visiting me:) about things you may want to check/look at:
- check that last page restoration works, and that you stay on the page you were after changeing some rendering settings (we may have changed/optimized some stuff)
- one thing we have optimized is not reparsing the stylesheet on each page turn, but you might need that. Implemented in https://github.com/koreader/crengine/pull/218, initial thoughs in https://github.com/koreader/crengine/pull/164#issuecomment-384877976:
> I believe we could get rid of this, as it looks to me it's for CoolReader behaviour: with CoolReader, we may change from the UI various settings, including styles, so that's crengine way of knowing some things have changed and that the rendering needs to take them into account.
With KOReader, as for each typeset/layout change, we are explicitely calling methods on crengine to set them, before rendering, we may not need that section.

- in KOReader, we always have only one single CRE document loaded. This has allowed me to take shortcuts and use some cheap global variables for some settings (global render dpi for sizing CSS https://github.com/koreader/crengine/pull/213, global interline scale factor https://github.com/koreader/crengine/pull/273...). I believe CoolReader can have multiple documents opened, and so, changing these settings on a document will have it changed for all other opened documents, meaning some style hash change that will cause a re-rendering and cache update.
I think this was already the case for the floating/hanging punctuation toggle that was there before I joined in, as you have that `bool gFlgFloatingPunctuationEnabled`. So, if you can toggle that from your frontends' UI, you should meet the situation I describe. Up to you to see if that's an issue or not.
Might not be an issue if you hardcode one value and don't allow for frontends to tune it, but well :)

pkb:
> I believe CoolReader can have multiple documents opened
No, it can switch to recently opened document but that's all.  CoolReader is a single document app.

poire-z:
> No, it can switch to recently opened document but that's all. CoolReader is a single document app.

Oh, allright. The lvtinydom code allows for multiple document instances. And I just remember we reduced the max number of such instances from 256 to 16, to allow for more nodes and bigger books (https://github.com/koreader/crengine/pull/121#issuecomment-457244709, https://github.com/koreader/crengine/pull/254).

pkb:
> PROP_RENDER_BLOCK_RENDERING_FLAGS "crengine.render.block.rendering.flags" - which should default to 0 (legacy rendering), 

In the lvrend.h #define DEF_RENDER_BLOCK_RENDERING_FLAGS BLOCK_RENDERING_FULL_FEATURED

Do I need to change it to 0 to turn all enhancements off? If in the future we will add UI allowing to turn these features on, user will still lose their bookmarks, right?

crengine/src/hist.cpp
```diff
@@ -516,6 +526,7 @@ ldomXPointer CRFileHist::restorePosition( ldomDocument * doc, lString16 fpathnam
     int index = findEntry( name, path, (lvsize_t)sz );
     if ( index>=0 ) {
         makeTop( index );
+        gDOMVersionRequested = _records[0]->getDOMversion();
         return doc->createXPointer( _records[0]->getLastPos()->getStartPos() );
```
poire-z: I wonder if that is not too late (the DOM might already be built if you are calling doc->createXPointer(), and so that old saved value wouldn't have any effect ?)
Also, given that you skipped all the intermediate DOM_VERSIONs and we are hopefully done with that work and there won't be any new one :), once this is merged, you have only 2 cases to handle:
- books with old hist files without yet any dom_version saved: you should request a very old one (we use `20171225 -- arbitrary day in the past`), and save that to the hist file, so it's used with next openings
- newly opened books with no hist file: use the current dom version, so they get all the good/fixed dom stuff

Note: the need for that DOM version is only because some nodes are switched from/to `display: block` to/from `display: inline` - and this may cause some of them to be (or no more be) wrapped in `<autoBoxing>` elements (anonymous block container for inline nodes that were at the same level of other block nodes).
That might be rare with well-formed books - but it may happen with our change to list item handling (that are now a proper block element, while there were inline previously and may have been wrapped in a `<autoBoxing>`  if there were `<P>` siblings for example): so you may have chances to see and test for highlights survival or disappearance by highlighting text inside lists.

pkb:
> I wonder if that is not too late (the DOM might already be built if you are calling doc->createXPointer(), and so that old saved value wouldn't have any effect ?)

I'm afraid you are right it is too late.

poire-z:
> In the lvrend.h #define DEF_RENDER_BLOCK_RENDERING_FLAGS BLOCK_RENDERING_FULL_FEATURED
Do I need to change it to 0 to turn all enhancements off? If in the future if we will add UI allowing to turn these features on, user will still lose their bookmarks, right? If so, how do you handle this, do you warn them?

Oh, ok, I did set it to full featured by default :) Yes, you could set it to 0 to get back the "legacy" behaviour.
And/or provide it at init time from frontend via the lvdocviewprops PROD_.

Indeed, when you change rendering flags, or stylesheets, lvtinydom computes a hash with `tinyNodeCollection::calcStyleHash()` (that we have optimized to skip doing that all the time, not sure if that can affect you), and that hash includes all the `display:` values of all the nodes.
If that hash is different from the previous one, a full re-rendering happens (same as when you change font size...).
We store that sub-hash in a `_nodeDisplayStyleHash`, and our frontend calls `tinyNodeCollection:isBuiltDomStale()` (after each page update, style change...) to know if it has changed.
If it has, we indeed warn them. We do that in https://github.com/koreader/koreader/blob/c8b3942bf45ef70b4018b2d1d8272c19d8c0eb57/frontend/apps/reader/modules/readerrolling.lua#L226-L263

poire-z:
Referencing our main post about that DOM change & breaking highlights: https://github.com/koreader/koreader/issues/3940.

There's was another possibility mentionned: making old xpointers work with a different DOM - because the different DOM is just different in that there are internal/shadow elements inserted or removed (autoBoxing, floatBox, inlineBox), and may be the ldomXPointer code could be made clever to work out what has changed itself, or walk a DOM tree/count siblings no matter what shadow elements are wrapping them.
I thought about that at https://github.com/koreader/koreader/issues/3940#issuecomment-388319566 - but I was naively convinced at the time that is was not possible.
But may be it is, it's more of a mathematical/theoritical problem :) and I suck at that. So, if one of you doesn't suck at maths and walking trees theory, it might be worth investigating if it could just be solved there, instead of relying on that cre_dom_version (which I'm not super sure I always updated, or if there are cases that may happen for other reasons that the changes I've been aware they could cause these issues...). It would be nice if that was possible :)

pkb:
> Oh, allright. The lvtinydom code allows for multiple document instances. And I just remember we reduced the max number of such instances from 256 to 16, to allow for more nodes and bigger books ([koreader/crengine#121 (comment)](https://github.com/koreader/crengine/pull/121#issuecomment-457244709), [koreader/crengine#254](https://github.com/koreader/crengine/pull/254)).

I stand corrected, @buggins said CoolReader keeps a document open when they are switched, thus CoolReader can have multiple documents opened

pkb:
The only real problem I see in your current approach of keeping compatibility is in difficulty to maintain it. It requires your level of competence and patience. I will definitely break something, (fortunately I'm not much involved in the project). On the other hand even if we will fix somehow an issue with automatically inserted nodes I might wish to change the DOM rather radically and in this case your compatibility mechanism is the good thing to have.

poire-z:
@pkb : you might want to at least pick https://github.com/koreader/crengine/commit/1bbc4d942ef28e302a1e24693f350c69e4e005d4 (from https://github.com/koreader/crengine/pull/326): for something I missed in the `Better selection highlighting by using getSegmentRects()` that you already picked - to not have mixed kinds of highlights.

NiLuJe:
As for the image rendering changes related to the scaling changes, we only ever tested 8bpp grayscale and 32bpp **bgr**. Stuff might be horribly broken in other cases (CRe has a *lot* of legacy handling of really old platforms with funky fb setup).

poire-z:
How will the xpointers look when converted using your new usingIndexes ? like `[2][4][6].4` ?
Are the converted xpointers understandable by older CoolReader versions? In case people are not satisfied with the newer one, will they be able to downgrade and find their previous bookmarks working?

Also, after a quick read of your code, it looks like you're not recursively walking down Boxing nodes, but just allow for one level?
That might not be an issue when migrating from current CR that has only autoBoxing, as I think an autoBoxing can not contain another autoBoxing,
But in principle, an autoBoxing can contain a floatBox or a inlineBox.
```html
<div>
  aaaa aaa
  <span style="display: block">
    bbbb bbb
      <span style="display: block">
        cccc ccc
      </span>
    dddd ddd
  </span>
  eeee eee
  <span style="float: left">
    <div>ffff fff</div>
    gggg ggg
  </span>
  <span style="display: inline-block">
    <div>hhhh hhh</div>
    iiii iii
  </span>
  jjjj jjj
</div>
```
<kbd>image</kbd>
<kbd>image</kbd>

pkb:
@poire-z, thanks for the example. it is very helpful.  I indeed did not account for auto boxing in more than one level. The XPath will look like this:

``<start-point>/1/2/4/4/1/1.0</start-point>
``

Yes, it doesn't look very readable (for a human). But I guess it should work, considering that number of nodes actually do not change as a result of boxing. Just structure changes.

This XPath syntax is supported by older versions of CRe but of course they do not take into account auto boxing, so they actually won't work (for some cases).

Unlike you I don't care much about users :), but @buggins should I guess.
For Qt I will not bother about possibility to downgrade and keep bookmarks working. If they care about theirs bookmarks, they can make backup. About Android I'll leave it to @buggins and @virxkane to decide how to proceed with this PR and the particular change in XPath format.
Will they ask user for permission to update bookmarks or just preserve old format along with the new one I don't know, they might even reject this PR and this change will not go anywhere

poire-z:
For debugging, I'd probably miss the xpath syntax with element names :)
But I wonder if your number-only syntax could not be used just internally.
With KOReader, the xpaths reach and are stored on the frontend side. With CR, I guess it's all managed internally by hist.cpp? Do you rewrite all the bookmarks when quitting the book?

If older and newer bookmarks were kept and saved with the name syntax, I guess they can be converted unambigously to the by-number syntax on load and these will be used internally for the session. On quit, leave the older names as-is, save the new ones by name with the new DOM structure. Users downgrading would see their older unchanged bookmarks, but not the newer ones - which would be acceptable I guess :)

Or even just don't convert at all: just have your code for skipping boxing nodes used internally when looking up a ldomNode from a xpath string (so, it would work with old and new xpaths) - and for the reverse, ldomNode > string, generate it with names with the current DOM structure.
Would that work?

poire-z:
> if one of you doesn't suck at maths and walking trees theory

I guess you now have thought about it :) so are you convinced that for a same HTML document (that would always give the same non-boxed DOM tree), there is always an unambiguous 1-to-1 relationship between
- a ldomNode in a DOM tree boxed with some method A
- a xpointer saved from a DOM tree boxed with some method B
?

poire-z:
> they might even reject this PR and this change will not go anywhere

Well, of course it needs thinking, but don't give up (you and them) ! :) With your weeks of work on this PR and my 2 years of work on crengine, CR would be upgraded to the best low-resources EPUB renderer out there :)

I must still say that I only targetted EPUB/HTML, and tried to conform to CSS/HTML specs - and that I never tested some of the original formats CR was made for, and your russian readers might still use, like FB2 or the plain-text based ones (the "lib.ru" tweaks in the code). I tried to not break them when I understand a bit of the tweaks, but I can't promise I haven't :) so please test these.
But for EPUB/HTML, I'm quite sure we're fine and solid - and you get 2 years worth of testing with KOReader and robustness :)

But your choice: may be CR users wouldn't need all that. Even if floats & inline-boxes, and the Enhanced rendering stuff, are quite rare in current ebooks, and might be overkill for casual reading, the [list item change ](https://github.com/koreader/crengine/pull/188) is quite a necessity with some books.
Also see https://github.com/koreader/crengine/pull/194#issuecomment-392834289 - I think you might still need to update epub.css to get correct EPUB rendering - and a more general thinking on your CSS (and maybe your styling UI).
The real issue is the xpath/xpointers stuff mentionned above: if this can really be made unambiguously 1-to-1, that would allow for future incremental changes and tweaks - without the need to have all that good and decided in a single PR.

pkb:
> But I wonder if your number-only syntax could not be used just internally.

It seems we can avoid using this syntax and return to identifiers. if we do everything correctly the XPath should match original tree, for your example pointer to gggg will be:

```
        <start-point>/html/body/div/span[2]/text().1</start-point>
        <end-point>/html/body/div/span[2]/text().5</end-point>
```

> With KOReader, the xpaths reach and are stored on the frontend side. With CR, I guess it's all managed internally by hist.cpp? Do you rewrite all the bookmarks when quitting the book?

For Qt and maybe some other frontends the answer is yes to both questions. On Android CR uses sqlite3 DB. 

> I guess you now have thought about it :) so are you convinced that for a same HTML document (that would always give the same non-boxed DOM tree), there is always an unambiguous 1-to-1 relationship 

I'm a programmer not a mathematician :) so no guaranties. It is as usual it works until it doesn't and then we fix it, if we can :). The same story as with your compatibility efforts. You do your best to provide it but you can't be 100% sure if something haven't broken.

> I think you might still need to update epub.css to get correct EPUB rendering - and a more general thinking on your CSS (and maybe your styling UI).

We will update epub.css before this PR will be merged into master. About UI, I think @virxkane will have to think about it :)

poire-z:
> The same story as with your compatibility efforts. You do your best to provide it but you can't be 100% sure if something haven't broken.

Except that in my case, all mathematicians (and non-mathematicians) in the world will agree I'll have failed at some point :) while for xpaths/tree, there must be some strong theory for 100% ok or for 0% :)

But ok, sharing more thoughts about these xpointers.
(I haven't really gotten into your code, my mind is on some other topics - so I may be wrong with my quick reading.)

Let's call your new xpointers, that don't include boxing element names, that map to a virtual non-boxed DOM tree (virtual for crengine, but the real HTML one): "normalized xpointers".
And call the past xpointers that were made before and that may include boxing element names: "legacy xpointers".

These normalized xpointers would always map to the same real ldomNode, as your new code would walk boxing Nodes (that may change depending on code change or CSS with different `display:`) to find the real node.
We can't translate a legacy xpointer directly to a normalized xpointer: we need to go navigate the DOM.
For example:
```
/html/body/div[1]/text()[1].0
   if it is preceded by
/html/body/autoBoxing[1]/div[1]/
   is indeed really the 2nd DIV in BODY, so its normalized xpointer would be:
/html/body/div[2]/text()[1].0
```
So, may be, if we had the convention of having normalized xpointers starting with 2 `//`:
`//html/body/div[2]/text()[1].0`
we could keep legacy xpointers (without a single leading `/`) and avoid any migration.
createXPointer() would return only normalized xpointers, but we could still use legacy xpointers to reference nodes: where we use:
`ldomXPointer startp = dv->createXPointer(lString16(pos0));`
would accept old or new pos0, but would return a normalize xpointer.
I guess createXPointer(lString16) would need 2 methods to walk nodes, the one for legacy xpointers being more costly? Or would they be and cost the same?

If your walking code would work the same with both, I guess then everything will be now a bit more expensive?
If you get `//html/body/div[245]/text()[1].0`, do you still need to go look at BODY 400 children, preceeding siblings of the 245th DIV, to see if any is a autoBoxing, and count the numbers of DIV inside each of them?
So, for a normalize xpointer:
`//html/body/div[245]/div[126]/ul[41]/li[26]/text()[1].0`,
it would make many more iterations than a straight legacy xpointer:
`/html/body/autoBoxing[15]div[115]/autoBoxing[12]/div[44]/ul[41]/floatBox[26]/li[1]/text()[1].0`,
if both map to the same node (the 1st one in the virtual DOM tree, the 2nd one in the real internal crengine DOM tree). Right? I hope that cost is negligeable.

(I'm a bit confused writting all this without having really studied how your code really do :) just some thoughts to hopefully  get yours straight and clear :)

poire-z:
Or may be I get it: for a legacy xpointer, you actually need the correct version legacy-ly built DOM tree to be able to go to the node? As with legacy `/html/body/div[1]/text()[1].0`, you actually don't know how many preceeding DIV were moved in how many autoBoxing - if you don't have the DOM tree if comes from?

pkb:
Yes, I'm not trying to make some super clever code that would resolve Xpointer composed with arbitrary transformation to the underlying document node. I assume, we have still working bookmarks, thanks to your compatibility code. Having this node, We make this normalized Xpointer as you call it. And with this pointer we are avoiding issues with tree transformations because of css changes.

poire-z:
> @virxkane in #126 : _can you upload your epub test files to test the new features that we got_ 

Here's [DocumentsForTestingRendering.zip](https://github.com/buggins/coolreader/files/4121240/DocumentsForTestingRendering.zip) with 3 Koreader-made Wikipedia EPUBs (with floats, inline block, tables, multi-level list items), as well as my big test-float-misc.html with one year of test cases :) (the text are not always accurate, so don't trust them).

A few more notes:
- about `DEF_RENDER_BLOCK_RENDERING_FLAGS` which is set to `BLOCK_RENDERING_FULL_FEATURED`, if you're not yet allowing users to change them, it would be best to go with something a little less full featured :) Images and stuff can overflow the screen when full featured, while disabling some of the features might make things more safe. Ref https://github.com/koreader/crengine/pull/299
In KOReader, we provide 4 sets, see: https://github.com/koreader/koreader/blob/5724971466bd8380fc1856b4556894b30ff74fd4/frontend/apps/reader/modules/readertypeset.lua#L321-L353. I'd go with the 3rd, "book", which have floats but limits overflow.
(The Wikipedia EPUBs uses [some private `-cr-only-if:` CSS properties](https://github.com/koreader/crengine/pull/299/commits/6a323526dfd7638738d5098aee5d6c4e7c5f731d#diff-1949e62eb9de850feaada4f91fd08fb2) to allow us to have 3 different looks whether flat|book|web.)

- I haven't tested how/if CR "image scaling/zoom" tunables work, so please do:). We don't use them in KOReader, so we are probably using the default values. We do scale images with https://github.com/koreader/crengine/pull/213 (`gRenderDPI / DEF_RENDER_DPI 96`).

- @pkb : _We will update epub.css_
would be nice, but I think your xpointer migration code would need the old DOM to be built with the same old epub.css.

- I tested cr3.2.39 (haden't upgraded CR since some cr3.1.2-33 from years ago) on my phone, and it renders quite fine. My only surprise when testing some Wikipedia EPUB was that the left margin was larger than the right margin by default, so quite unbalanced (might be from your epub.css).

pkb:
@poire-z you are spot on. We'll have to keep this epub.css and if we will need to modify css for epub we will have to create new file. I hate this compatibility stuff :).

poire-z:
^ but I had forgotten since last week :| and I just remember that, maybe, this can be taken care of with:
> This also affected our epub.css, that has this ugly section at its bottom:
https://github.com/koreader/crengine/blob/f20454111fff249d78a1f092ff7ccba52c8b720f/cr3gui/data/epub.css#L170-L207

(you might want to check our old epub.css vs your current one, could be there's been changes before I jumped in, and you need to adapt it a bit if you're goind with these tricks)

virxkane:
@poire-z Thank you for file DocumentsForTestingRendering.zip. Can you upload some epub file with RTL text to for testing purposes?

poire-z:
Here: [DocumentsForTestingRTL.zip](https://github.com/buggins/coolreader/files/4132765/DocumentsForTestingRTL.zip) (including some saved html from https://r12a.github.io/scripts/phrases)

(Also note https://github.com/virxkane/coolreader/commit/8975b2c1ce47052da0b7a3b0dc41c4a2b84c1386#r37034546 about FLAT and not FLOAT)

Another minor thing 
```diff
--- ko-crengine/crengine/include/hyphman.h      2020-01-20 07:21:47.903389779 +0100
+++ pkb-coolreader/crengine/include/hyphman.h   2020-01-18 09:27:45.629120980 +0100
-#define WORD_LENGTH   64
-#define MAX_REAL_WORD 24
+#define WORD_LENGTH   2048
+//#define MAX_REAL_WORD 24
```
The 2048 you kept shouldn't really need to be that large. (and recently, I added in lvtextfm.cpp `#define MAX_WORD_SIZE 64` that should really have been the same - but no issue if you have it larger, just a bit of waste :)

pkb:
I agree 2048 is overkill :) and the way it is used now it is indeed waste.  RTL requires freebidi, right?

poire-z:
> RTL requires fribidi, right?

Right.

About my comment:
> My only surprise when testing some Wikipedia EPUB was that the left margin was larger than the right margin by default, so quite unbalanced (might be from your epub.css).

This issue is related to Floating punctuation being enabled (your default). When I disable it, the unbalance is still there, but if I close and re-open the document, it's then all fine and balanced.

Note that there are quite ugly issues with how Floating punctuation, that I haven't fixed (because it's mostly some code related to CJK/Chinese characters, and I can't check if I would made thing better or worse as I don't read the languages - but I'll probably have a go at it some day). See https://github.com/koreader/koreader/issues/2844.
On KOReader, we switch some months ago the default for Floating/hanging punctuation to disabled: it's safer for western readers and new users.

Also, when testing your cr3.2.39 apk with a big book (not free, can't share, but any christian Bible would do I guess :), page turn is really slow (5 to 10 seconds, might be just node styles hash recomputing, or a hidden reformating), and reformatting (with progress bar) happen on load (so, like a cache hash mismatch).
So, you might have some issue with hash/style hash/display hash checks on page turn :( (I got that a few times, and it had always been painful to debug...) May be you'll have some `CRE WARNING:` in your output that could give some (limited) hints?
May be it's related to your stylesheet injection/check on each page turn, and the optimisation I mentionned in the 2nd point of https://github.com/buggins/coolreader/pull/125#issuecomment-576240477. Or not.

(I do think your current code before this merge did have some cache/hash bugs that would cause some reparsing/reformatting on load - but you should have benefit from my fixes to that with this merge - except that your style injection is different than ours, and it may bring some new issues...)

poire-z:
For info (I just stumbed on it when looking at fixing an unrelated koreader issue):
https://github.com/koreader/koreader/pull/3740 we also did fix a bit of page turn slowness on our frontend side.
But I don't think it's related to the slow page turn i described above, as setting `Highlight bookmarks` to `None` in your android options (which I guess set `crengine.highlight.bookmarks=0`) did not change the slowness.

pkb:
Thanks @poire-z, it is good this was discovered before release. Page turn slowness is not critical we can deal with it later, but Floating punctuation, we need to figure something out before the release. I remember how @buggins had to turn CJK patch off because of users complaints.

poire-z:
(Eyeing on your gitter with the help of google translate.)
@virxkane : don't try to fix floating punctuation in lvtextfm.cpp, it's quite messy and complicated. There will always be more indentation (but it should be balanced I think), as that's how it's supposed to work: it kinda adds some padding.
May be re-applying https://github.com/buggins/coolreader/commit/c159100630840095d3d53dad7e0db543a1bbf50c (may be also some other commits around) that you had for years (unlike us) before this merge would get you back the previous behaviour?

pkb:
I've introduced this margins issue. updatePageMargins() (which you don't have) is called few times and each time it performs this visual alignment for floating punctuation, making left margin bigger and the right one smaller.

poire-z:
@virxkane : regarding your 2 commits https://github.com/virxkane/coolreader/commit/29b2030cdaab347bf6224760728ec2e48daece3c and https://github.com/virxkane/coolreader/commit/0765ffe0f33d8d1e92b09a8313fceed8ed4d0e45
I understand the fix in the first is needed, as it uses `getChildElementNode(i)` which may return NULL if child is not an element.
But in the 2nd commit, it seems we are always using `getChildCount()` and then `getChildNode( i )`, which should always return a valid child (never NULL for `0 <=i < getChildCount()` - so all these seems uneeded (and if they were, there are many other places we use that that would need fixing :)
So, is that just extra security ("potential", but there's no potential :), or have you really witnessed a crash with that? Or is it just static analyzer warnings?

virxkane:
@poire-z 29b2030 fixes real crash. So I think if GetChildElementNode() can return NULL this good idea to check this in other places too. Later I will upload html file that crashed desktop version of CoolReader. Maybe 0765ffe is useless but I unsure, lvtinydom.cpp is too complicated for me. Introduced this crash I think (also by me) in https://github.com/buggins/coolreader/pull/56

virxkane:
Uploaded test file, unzip and open "China (ar).html". Crashed if view mode is "page". Stack trace on screenshop.
[China (ar).zip](https://github.com/buggins/coolreader/files/4146339/China.ar.zip)
Screenshot_20200203_081932

virxkane:
Also second commit (https://github.com/virxkane/coolreader/commit/0765ffe0f33d8d1e92b09a8313fceed8ed4d0e45) I hope fixes other segfault with stacktrace below (version 3.2.36):
```
#00 pc 00000000000e6878 operator[] at crengine/include/lvptrvec.h:46
#01 pc 00000000001e0df0 renderBlockElement(LVRendPageContext&, ldomNode*, int, int, int) at crengine/src/lvrend.cpp:1427
#02 pc 00000000001e1b9c renderBlockElement(LVRendPageContext&, ldomNode*, int, int, int) at crengine/src/lvrend.cpp:1513
#03 pc 00000000001e1b9c renderBlockElement(LVRendPageContext&, ldomNode*, int, int, int) at crengine/src/lvrend.cpp:1513
#04 pc 00000000001e1b9c renderBlockElement(LVRendPageContext&, ldomNode*, int, int, int) at crengine/src/lvrend.cpp:1513
#05 pc 00000000001e1b9c renderBlockElement(LVRendPageContext&, ldomNode*, int, int, int) at crengine/src/lvrend.cpp:1513
#06 pc 00000000001e1b9c renderBlockElement(LVRendPageContext&, ldomNode*, int, int, int) at crengine/src/lvrend.cpp:1513
#07 pc 00000000001e1b9c renderBlockElement(LVRendPageContext&, ldomNode*, int, int, int) at crengine/src/lvrend.cpp:1513
#08 pc 00000000000e2934 ldomDocument::render(LVRendPageList*, LVDocViewCallback*, int, int, bool, int, LVProtectedFastRef<LVFont>, int, LVFastRef<CRPropAccessor>) at crengine/src/lvtinydom.cpp:3384
#09 pc 00000000001b74e4 LVDocView::Render(int, int, LVRendPageList*) at crengine/src/lvdocview.cpp:2575
#10 pc 00000000001b71e0 LVDocView::checkRender() at crengine/src/lvdocview.cpp:595
#11 pc 00000000001b9480 LVDocView::checkPos() at crengine/src/lvdocview.cpp:605
#12 pc 00000000000a5da8 Java_org_coolreader_crengine_DocView_getPositionPropsInternal at android/jni/docview.cpp:1638
```
But I unsure, coredump ommited.

poire-z:
OK, thanks.
I'll pick the first one for KOReader. But I think I'll let the 2nd out, as what it checks for shouldn't happen - so, it may mask a deeper bug that should probably be fixed in some other bit of the code.

Also, given the amount of changes you got with 3.2.39, including my part of segfault fixes :), it's possible bugs from 3.2.36 got fixed - and I wouldn't spend time on older crengine related bugs.
And when you get new bugs or segfault reported, have a check on our commits since your last sync/picks, they might have already been fixed on our side :)

crengine/src/lvdocview.cpp
```diff
@@ -3798,13 +3799,25 @@ bool LVDocView::LoadDocument(const lChar16 * fname, bool metadataOnly) {
 				(int) stream->GetSize()));
 		m_doc_props->setString(DOC_PROP_FILE_NAME, arcItemPathName);
         m_doc_props->setHex(DOC_PROP_FILE_CRC32, stream->getcrc32());
-        const CRFileHistRecord* record = m_hist.getRecord( filename16, stream->GetSize() );
-        gDOMVersionRequested = record ? record->getDOMversion() : gDOMVersionCurrent;
+        CRFileHistRecord* record = m_hist.getRecord( filename16, stream->GetSize() );
+
+        if(record) {
+            gDOMVersionRequested = record->getDOMversion();
+            m_doc_props->setInt(PROP_RENDER_BLOCK_RENDERING_FLAGS, 0);
```
virxkane: This is typo? ```m_doc_props->setInt(PROP_RENDER_BLOCK_RENDERING_FLAGS, 0);```
May be must be ```m_props->setInt(PROP_RENDER_BLOCK_RENDERING_FLAGS, 0);```

crengine/src/lvdocview.cpp
```diff
@@ -3798,13 +3799,25 @@ bool LVDocView::LoadDocument(const lChar16 * fname, bool metadataOnly) {
 				(int) stream->GetSize()));
 		m_doc_props->setString(DOC_PROP_FILE_NAME, arcItemPathName);
         m_doc_props->setHex(DOC_PROP_FILE_CRC32, stream->getcrc32());
-        const CRFileHistRecord* record = m_hist.getRecord( filename16, stream->GetSize() );
```
virxkane: Possible NULL pointer dereference. getRecord() can return NULL.

crengine/src/lvdocview.cpp
pkb: Sure, it should be m_props.

crengine/src/lvdocview.cpp
pkb: convertBookmarks flag will be false when record is NULL.

virxkane: Ok, you right.

crengine/src/lvdocview.cpp
```diff
@@ -3542,8 +3748,21 @@ static void FileToArcProps(CRPropRef props) {
...
+static bool needToConvertBookmarks(CRFileHistRecord* historyRecord)
+{
+    bool convertBookmarks = false;
+    if(historyRecord) {
+        gDOMVersionRequested = historyRecord->getDOMversion();
+        if(gDOMVersionRequested < 20180528) {
+            convertBookmarks = historyRecord->getBookmarks().length() > 1;
+        }
+    } else
+        gDOMVersionRequested = gDOMVersionCurrent;
```
poire-z: Been wondering how much of that CRFileHistRecord stuff in that LVDocView::LoadDocument() I can keep without risk for KOReader, as we don't use CRFileHistRecord, so all that stuff should be no-op.
Is that the case too with CR on Android?
But this line bothers me: `needToConvertBookmarks()` is called even if there's no CRFileHistRecord (meaning all that is managed from outside), so it looks like this `else` is wrongly killing the gDOMVersionRequested the outside code may have requested?

pkb: If we change needToConvertBookmarks() like below:

```
diff --git a/crengine/src/lvdocview.cpp b/crengine/src/lvdocview.cpp
index d2ea6dc..ff7928e 100644
--- a/crengine/src/lvdocview.cpp
+++ b/crengine/src/lvdocview.cpp
@@ -3751,13 +3751,12 @@ static void FileToArcProps(CRPropRef props) {
 static bool needToConvertBookmarks(CRFileHistRecord* historyRecord)
 {
     bool convertBookmarks = false;
-    if(historyRecord) {
+    if(historyRecord && historyRecord->getBookmarks().length() > 1) {
         gDOMVersionRequested = historyRecord->getDOMversion();
         if(gDOMVersionRequested < 20180528) {
-            convertBookmarks = historyRecord->getBookmarks().length() > 1;
+            convertBookmarks = true;
         }
-    } else
-        gDOMVersionRequested = gDOMVersionCurrent;
+    }
     return convertBookmarks;
 }
```

This way it will not affect you, right?

poire-z: That looks fine. (So, that way, you're not resetting at all gDOMVersionRequested , but it will be gDOMVersionCurrent when you have no historyRecord because... you set/reset it to gDOMVersionCurrent elsewhere?)

pkb: Initially gDOMVersionRequested is equal to gDOMVersionCurrent. I did not find were is it changed. I assume it is changed in a frontend or I missed something. So the one who changes gDOMVersionRequested  is responsible for resetting it.

I did some more changes, please take a look: https://gist.github.com/pkb/64714978e3a467b687b738c3a0f3b51b

Does it make sense?

poire-z: > So the one who changes gDOMVersionRequested is responsible for resetting it.

OK, In KOReader, we set it from frontend before each book loading (from the value saved in our book settings, or to the gDOMVersionCurrent that we have fetched from CRE, when the book has no setting yet).
I get that in your case, for each book, you set it first to the value saved in your CRHistFile (so, the oldest if none), always migrate bookmarks and always end up setting it to gDOMVersionCurrent. So I guess both you and me are fine (as long as you don't tweak gDOMVersionRequested when there is no CRHistFile :)
Your patch looks sane to me, when reading it in your and my contexts.

But I don't know how that work for CR Android, You told me bookmarks were saved in a SQLite DB, so I assume it's not using CRHistFile, right?. Is that gDOMVersion business and bookmarks migration taken care in any way on CR Android?

Also, a question, just because I don't know much about CRHistFile: is it always fed from outside frontends? So, if that is not done by the frontend, it's always NULL and checks with it are just skipped, right?

Btw, in your merged code and patch, you might want to use (instead of your hardcoded `< 20180528`):
```c
#define DOM_VERSION_WITH_XPOINTERS_V2 20180528
or
#define DOM_VERSION_WITH_NORMALIZED_XPOINTERS 20180528
```
(because for me, that won't be `20180528`, but something like `20200214`, depends on when I'll get that ready :) or you could have used anything 2020xyxz that I could reuse - but minor diffs, no problem).

`historyRecord->getBookmarks().length() > 1` - another (less critical) thingy that might need conversion is the last position in book. After migration and switch to the latest DOMVersion, the last position might not resolve - unless crengine handles that somehow (I've seen it doing that in some other contexts, like re-renderings, but not sure if it uses a xpointer or a y-coordinate).

(In CRFileHistRecord::convertBookmarks(), your toggling of gDOMVersionRequested  between old and new versions is just to pick the right version of createXPointer() and toString(), right? would have worked the same by using createXPointerV1 and toStringUsingNames(New), right?)

pkb: I don't know much about Android part. Let's wait and see what @virxkane has to say about it.
AFAIU new CRFileHistRecord gets created whenever LVDocView::savePosition() is called. So you might have a record if you call LVDocView::goLink() for example. It will not be saved to a file, so on start record list will be empty. But anyway getBookmarks().length() will be at most 1 so 

> Btw, in your merged code and patch, you might want to use

Ok

> another (less critical) thingy that might need conversion is the last position in book

Yes I'm aware of this. Maybe I'm wrong but I've decided to skip this conversion. There is a chance that reading position will be lost.

> In CRFileHistRecord::convertBookmarks(), your toggling of gDOMVersionRequested

Right.

@virxkane could you please check the patch https://gist.github.com/pkb/64714978e3a467b687b738c3a0f3b51b and apply it along with your other changes if you find it ok? Thanks

virxkane: @pkb Ok, it's done. https://github.com/virxkane/coolreader/commit/999a484ee1f9c4a9287932537fae9697f96f7d28
@poire-z About your question: yes, on Android bookmarks saved in SQLite DB, so CRHistFile is unused. Bookmark migration on Android for now not implemented yet. In latest commit https://github.com/virxkane/coolreader/commit/3189393b1067ddab0b33375ebbf1f40014258edb I implement saving & loading DOM version of each document in DB, so at least old bookmark must be handled properly. In the future, the conversion of bookmarks will probably be implemented.

poire-z: Some update from my side: I brought these normalize xpointers in https://github.com/koreader/crengine/pull/328 - with some additional minor tweaks and renames.

> AFAIU new CRFileHistRecord gets created whenever LVDocView::savePosition() is called. So you might have a record if you call LVDocView::goLink() for example.

Ok, we're indeed using that, only when resizing or changing orientations. Didn't take any risk and let your CRHistFile conversion code out.

poire-z:
Another thing I forgot and that I should mention:
The default caches size in lvtinydom.cpp are really small. I did not dare updating them in the source file, but I added stuff so frontend code can feed a factor to have them get bigger. We currently pass a factor of 20, and I'm going to update it to 40.
Reaching the limits on these caches is the cause of other obscure bugs and slowness. Details in:
https://github.com/koreader/crengine/pull/100 - https://github.com/koreader/koreader/pull/3700 - https://github.com/koreader/koreader/pull/3732

virxkane:
> Also, when testing your cr3.2.39 apk with a big book (not free, can't share, but any christian Bible would do I guess :), page turn is really slow (5 to 10 seconds, might be just node styles hash recomputing, or a hidden reformating), and reformatting (with progress bar) happen on load (so, like a cache hash mismatch).
> So, you might have some issue with hash/style hash/display hash checks on page turn :( (I got that a few times, and it had always been painful to debug...) May be you'll have some `CRE WARNING:` in your output that could give some (limited) hints?
> May be it's related to your stylesheet injection/check on each page turn, and the optimisation I mentionned in the 2nd point of [#125 (comment)](https://github.com/buggins/coolreader/pull/125#issuecomment-576240477). Or not.
> 
> (I do think your current code before this merge did have some cache/hash bugs that would cause some reparsing/reformatting on load - but you should have benefit from my fixes to that with this merge - except that your style injection is different than ours, and it may bring some new issues...)

I found root of this problem. I think it's affected only Android version. Method ```org.coolreader.crengine.DocView.getPositionPropsInternal()``` call ```LVDocView::getBookmark()``` & ```LVDocView::getPageDocumentRange(int)```, which your team greatly modified and they became really really slower.
LVDocView::getPageDocumentRange(int):
```
    ...
    for (start_h=0; start_h < height; start_h++) {
        start = m_doc->createXPointer(lvPoint(0, start_y + start_h), 1);
        // printf("  start (%d=%d): %s\n", start_h, start_y + start_h, UnicodeToLocal(start.toString()).c_str());
        if (!start.isNull()) {
            // Check what we got is really in current page
            lvPoint pt = start.toPoint();
            // printf("start pt.y %d start_y %d end_y %d\n", pt.y, start_y, end_y);
            if (pt.y >= start_y && pt.y <= end_y) {
                found = true;
                break; // we found our start of page xpointer
            }
        }
    }
    ...
```
LVDocView::getBookmark()
```
	...
	for (int y = page->start; y < page->start + page->height; y++) {
		// Use direction=1 to avoid any x check (in case there
		// is some left margin)
		ptr = m_doc->createXPointer(lvPoint(0, y), 1);
		lvPoint pt = ptr.toPoint();
		if (pt.y >= page->start) {
			if (!fallback_ptr)
				fallback_ptr = ptr;
			if ( pt.y < page->start + page->height ) {
				// valid, resolves back to same page
				found = true;
				break;
			}
		}
	}
	...
```
Are these cycles really useful and necessary? I tried to increase the increment of the loop counter to 10, 100, it works much faster. But there is a risk of not finding anything at the end of the desired interval... What is the preferred increment value of the loop counter in your opinion? Or may be use Bisection method? Or something else?

poire-z:
> Are these cycles really useful and necessary? 

Well, usually, the for loops should not go on much, there's always a break that should get out of it really early.
It was needed when there is some top margin or some space at top, that would not resolve to any node: so, when it does not, try top+1px top+2ps... hopefully, top + a few pixels should resolve to a node, and we get out of it. So, really, the increment should be 1. If you jump by too large, you may miss some nodes (which doesn't matter for getBookmarks, but it may with getDocumentRage (full text search).
Some link to past issues at the bottom of the last PR about it: https://github.com/koreader/crengine/pull/313

But also see https://github.com/buggins/coolreader/pull/125#issuecomment-590080631 just above - and try to increase these cache sizes. It's possible these affect createXPointer(lvPoint) as it needs to go thru many RenderRectAccessor - and if they can't get cached, there might be some recomputing that would cause some slowdown.

> I think it's affected only Android version. Method org.coolreader.crengine.DocView.getPositionPropsInternal() call LVDocView::getBookmark() & LVDocView::getPageDocumentRange(int)

Can't really find out how that works, and where you need that and how often you call it... Android source tree is quite obscure to me :)
So, why/how often?

virxkane:
> Can't really find out how that works, and where you need that and how often you call it... Android source tree is quite obscure to me :)
> So, why/how often?

On a smartphone with page animation turned on - 9 times per page turn.
```
(java code) ReaderView.animatePageFlip() -> 
  (java code) DocView.getPositionProps() -> 
    (cpp  code) DocView.getPositionPropsInternal()
```
```
(java code) ReaderView.animatePageFlip() ->
  (java code) ReaderView.PageViewAnimation class -> 
    (java code) ReaderView.preparePageImage() ->
      (java code) DocView.getPositionProps() -> 
        (cpp  code) DocView.getPositionPropsInternal()
```
The latter is executes few times.

On e-ink bookreader (without animation) - 2 times per page turn.
```
(java code) ReaderView.doEngineCommand() -> // cmd=DCMD_PAGEUP
  (java code) ReaderView.updateCurrentPositionStatus() ->
    (java code) DocView.getPositionProps() -> 
      (cpp  code) DocView.getPositionPropsInternal()
```
```
(java code) ReaderView.DrawPageTask class -> 
  (java code) ReaderView.preparePageImage() ->
    (java code) DocView.getPositionProps() -> 
      (cpp  code) DocView.getPositionPropsInternal()
```

Increasing maximum alowed cache size not fixed this slowness. But I found integer overflow in cache size calculations :)

poire-z:
For info, with KOReader, we do not use any of the CMD stuff.
I think a few times I made some changes, I noticed there would be needed also in some CMD part of the code - but as I couldn't test any change there, I didn't touch these... (and of course, I can't remember which :|)

Anyway, in your samples above, if for a page turn, wouldn't they return always the same value, and would benefit being cached locally so you don't call them 9 times?

poire-z:
> Increasing the maximum allowed cache size not fixed this slowness. But I found integer overflow in cache size calculations :)

Did your fix (https://github.com/virxkane/coolreader/commit/36299e6e14) and increasing cache size help with the slowness?

(I noticed some slowness in some page turn after I added a new feature, that resulted in the for loop iterating all over the height - it had to be fixed elsewhere https://github.com/koreader/crengine/pull/330/commits/1cec90fbc4 . So, if you really notice the for loops being done fully, it might be a bug - hopefully only on some specific pages of some specific book, but not all?)

virxkane:
> Did your fix ([virxkane@36299e6](https://github.com/virxkane/coolreader/commit/36299e6e14)) and increasing cache size help with the slowness?

No, it not helps. This commit for future increasing cache that has not yet been applied.

> hopefully only on some specific pages of some specific book, but not all?)

Yes, the slowdown is not on all books, I hope your fix corrects this behavior.
Unfortunately, there is not enough time, but today I’ll try to play with page turn some more.

poire-z:
> I hope your fix corrects this behavior.

Possibly not, that fix was for taller-than-page elements with `display: inline-block`,  which must be quite rare.
Anyway, if you have some specific book (and page reference) where you notice the slowdown, and you can upload that book, I can have a look (to at least see if that slowdown is noticable with KOReader, and if yes, I'll investigate).

virxkane:
@poire-z DocumentsForTestingRTL.zip that you upload about month ago, file Petra.AR.epub. Slowdown occurs when I turn pages around contents with numeric list.

virxkane:
@poire-z Forgot to add: with DOMVersion and BlockRenderingFlags both equals 0.

poire-z:
OK, took that epub from that zip file. With KOReader, and just adding:
```diff
ldomXPointer LVDocView::getBookmark() {
        LVLock lock(getMutex());
        checkPos();
        ldomXPointer ptr;
        if (m_doc) {
                if (isPageMode()) {
                        if (_page >= 0 && _page < m_pages.length()) {
                                // In some edge cases, getting the xpointer for y=m_pages[_page]->start
                                // could resolve back to the previous page. We need to check for that
                                // and increase y until we find a coherent one.
                                // (In the following, we always want to find the first logical word/char.)
                                LVRendPageInfo * page = m_pages[_page];
                                bool found = false;
                                ldomXPointer fallback_ptr;
                                for (int y = page->start; y < page->start + page->height; y++) {
+                                       printf("page.start = %d : y=%d\n", page->start, y);
                                        ptr = m_doc->createXPointer(lvPoint(0, y), PT_DIR_SCAN_FORWARD_LOGICAL_FIRST);
                                        lvPoint pt = ptr.toPoint();
```

In enhanced mode, I always see a single loop iteration (ok, twice because we must call this twice per page turn, one to go to, one to get the xpointer to restore to):
```
page.start = 24575 : y=24575
page.start = 24575 : y=24575
page.start = 25127 : y=25127
page.start = 25127 : y=25127
page.start = 25674 : y=25674
page.start = 25674 : y=25674
```
Still a single one on the TOC at start, and on the footnotes at end.

There's one page when I get 100 loops:
```
page.start = 15742 : y=15742
page.start = 15742 : y=15743
page.start = 15742 : y=15744
page.start = 15742 : y=15745
[...]
page.start = 15742 : y=15845
page.start = 15742 : y=15846
page.start = 15742 : y=15847
page.start = 15742 : y=15848
```
Confirmed with an other LTR EPUB that this happens when some float is split among 2 pages, on the 2nd page:
1st page:
<kbd>img1</kbd>
2nd page:
<kbd>img2</kbd>

It looks logical: when you ask what xpointer is at top of that 2nd page, it gives you the float (or the image inside the float) that starts on previous page, you resolved it back to a y, which happens to be smaller than the one you asked for, so it is bad: so, you continue the loop, until you pass 100px (the height of the 2nd part of the image on that 2nd page), and you finally reach some node that is on this page (y > the y you asked for), and it's fine.
(^ that's to show you the purpose of that loop :) and to refresh my memory :)

In "legacy" mode, I notice all is fine, in the TOC and footnotes, BUT not at end, after the footnotes, when there is a table: there, in that table, there are many loops (but 100 to 200 px only). But well, legacy mode :) not worth investigating.

So, I don't notice what you notice on page with numeric list (TOC at start or footnotes at end?).
Do you notice it on LTR EPUBs too? or only that RTL one?

Some possibly related commits I made that you haven't picked yet, that may be worth quick applying/testing to see if they solve your issue?
https://github.com/koreader/crengine/commit/7ed1ad5ebe5c03e32ed9d2e38b04b270617edbd9
https://github.com/koreader/crengine/commit/343066d153ab3f19547961d60a0656d14e0e8b2a
https://github.com/koreader/crengine/commit/204ab8ad3072a7d6694e6733068be6f18c7d805e

poire-z:
> Forgot to add: with DOMVersion and BlockRenderingFlags both equals 0.

Uh ! that's quite a huge forgetting :)

Well, at minima, DOMVersion should be >= 20180528 (and you do that migration automatically, no?) - so you get at least proper display for elements, and the most essential fixes. Not fixing it as having DOMVersion=20180528 is actually the fix to many things :)
BlockRenderingFlags = 0 = legacy mode, so, some possible minor issues with the table at the end that I noticed above :)

virxkane:
@poire-z Ok, I copied all your last commits via cherry-pick and slowdown are fixed now. Thank you.
Before:
```
preparePageImage(): get currpos: elaped=272
```
After:
```
preparePageImage(): get currpos: elaped=12
```
Logcat at page turn action with some debug noise:
```
03-12 23:48:12.250 2222-2222/org.coolreader V/cr3: G|rv| onKeyDown(92, KeyEvent { action=ACTION_DOWN, keyCode=KEYCODE_PAGE_UP, scanCode=0, metaState=0, flags=0x0, repeatCount=0, eventTime=3023151, downTime=3023151, deviceId=-2023406815, source=0x12345678 })
03-12 23:48:12.250 2222-2222/org.coolreader D/cr3: G|rv| onKeyUp: action PAGE_UP found for key 92
03-12 23:48:12.250 2222-2222/org.coolreader I/cr3: G|rv| On command DCMD_PAGEUP (1)
03-12 23:48:12.250 2222-2222/org.coolreader D/cr3: G|rv| doCommand(DCMD_PAGEUP, 1)
03-12 23:48:12.250 2222-2222/org.coolreader D/cr3: G|en| executing task
03-12 23:48:12.250 2222-2248/org.coolreader I/cr3: B|en| running task.work() org.coolreader.crengine.ReaderView$15
03-12 23:48:12.250 2222-2248/org.coolreader V/cr3eng: doCommand(102, 1)
03-12 23:48:12.250 2222-2248/org.coolreader V/cr3eng: getCurrentPageBookmarkInternal: calling getBookmark()
03-12 23:48:12.260 2222-2248/org.coolreader D/cr3eng: new Xpath: /body/DocFragment/body/div/div[2]/div/h2/text().0, old Xpath: /body/DocFragment/body/div/div[2]/div/h2/text().0
03-12 23:48:12.260 2222-2248/org.coolreader V/cr3eng: getBookmarkPosText() : getting position text
03-12 23:48:12.270 2222-2248/org.coolreader E/cr3eng: getPositionPropsInternal(): elapsed=0,0,0,6,3
03-12 23:48:12.270 2222-2248/org.coolreader I/cr3: B|en| exited task.work() org.coolreader.crengine.ReaderView$15. elapsed 17 ms.
03-12 23:48:12.270 2222-2222/org.coolreader I/cr3: G|en| running task.done() org.coolreader.crengine.ReaderView$15 in gui thread
03-12 23:48:12.270 2222-2222/org.coolreader V/cr3: G|rv| drawPage() : submitting DrawPageTask
03-12 23:48:12.270 2222-2222/org.coolreader V/cr3eng: getCurrentPageBookmarkInternal: calling getBookmark()
03-12 23:48:12.270 2222-2222/org.coolreader D/cr3eng: new Xpath: /body/DocFragment/body/div/div[2]/div/h2/text().0, old Xpath: /body/DocFragment/body/div/div[2]/div/h2/text().0
03-12 23:48:12.270 2222-2222/org.coolreader V/cr3eng: getBookmarkPosText() : getting position text
03-12 23:48:12.280 2222-2222/org.coolreader D/cr3: G|en| executing task DrawPageTask
03-12 23:48:12.280 2222-2222/org.coolreader V/cr3eng: getCurrentPageBookmarkInternal: calling getBookmark()
03-12 23:48:12.280 2222-2248/org.coolreader I/cr3: B|en| running task.work() org.coolreader.crengine.ReaderView$DrawPageTask
03-12 23:48:12.280 2222-2248/org.coolreader E/cr3: B|rv| DrawPageTask.work(1072,1448)
03-12 23:48:12.280 2222-2248/org.coolreader V/cr3: B|rv| preparePageImage( 0)
03-12 23:48:12.280 2222-2248/org.coolreader D/cr3: B|rv| BitmapFactory: bitmap released, used size = 0, free size=1
03-12 23:48:12.280 2222-2222/org.coolreader D/cr3eng: new Xpath: /body/DocFragment/body/div/div[2]/div/h2/text().0, old Xpath: /body/DocFragment/body/div/div[2]/div/h2/text().0
03-12 23:48:12.280 2222-2222/org.coolreader V/cr3eng: getBookmarkPosText() : getting position text
03-12 23:48:12.290 2222-2248/org.coolreader E/cr3eng: getPositionPropsInternal(): elapsed=0,1,0,2,3
03-12 23:48:12.290 2222-2248/org.coolreader E/cr3: B|rv| preparePageImage(): get currpos: elaped=12
03-12 23:48:12.290 2222-2248/org.coolreader V/cr3eng: getPageImageInternal entered : bpp=4
03-12 23:48:12.450 2222-2248/org.coolreader E/cr3: B|rv| DrawPageTask.work(): elapsed = 176
03-12 23:48:12.460 2222-2248/org.coolreader D/cr3: B|rv| doDraw() called
03-12 23:48:12.460 2222-2248/org.coolreader D/cr3: B|rv| onDraw() -- drawing page image
03-12 23:48:12.470 2222-2248/org.coolreader I/cr3: B|en| exited task.work() org.coolreader.crengine.ReaderView$DrawPageTask. elapsed 196 ms.
03-12 23:48:12.470 2222-2222/org.coolreader I/cr3: G|en| running task.done() org.coolreader.crengine.ReaderView$DrawPageTask in gui thread
```

virxkane:
Hmm, no, I found slowdown in other file - Petra.FR.epub
device-2020-03-12-235845
Firstly I go to page on screenshot, then goto next page and:
```
preparePageImage(): get currpos: elaped=973
```
Again:
```
preparePageImage(): get currpos: elaped=1965
```
As before, the reason is the same - LVDocView::getPageDocumentRange(int), LVDocView::getBookmark().

poire-z:
Can't reproduce on this one.
(Note that there is a large table on the page before - but it shouldn't apply on having a slowdown on going to next page - where I have only text...)

But even with a DOMVersion=20100101 and rend flags=0, I get the image centered. I have it off like you do when I switched from "web" to "legacy" (but koreader tells me the cache is invalid and I need to reload the book - and then I get the image centered).
So, may be clean up your cache files.

And why are you testing with DOMVersion=0/rendflags=0 ? :)

virxkane:
> And why are you testing with DOMVersion=0/rendflags=0 ? :)

No, DOMVersion=last, BlockRenderingFlags=max.
On koreader this page looks different?

virxkane:
Ok, after cleaning cache directory problem fixes. So, on our side incorrectly switched DOMVersion (option in settings dialog), but in other cases not specified, the slowdown may return. This can be a trap.
device-2020-03-13-002900

poire-z:
Yep, I have it like that ^ , with a proper float :)

poire-z:
> So, on our side incorrectly switched DOMVersion (option in settings dialog), but in other cases not specified, the slowdown may return. This can be a trap.

You picked https://github.com/virxkane/coolreader/commit/2aa255072d8e0b8d873759f7ae7454dfc4cf03c9 - so it should be automatically handled: cache trashed when dom version different.
Also see https://github.com/buggins/coolreader/pull/125#issuecomment-576371520 above, that I just made even more sensitive with https://github.com/virxkane/coolreader/commit/b165b56bc572e0348f6afc2e02b387b41f9bc392. You might want to call after each setting change: `include/lvtinydom.h isBuiltDomStale()` to either ask the user, or just reload the book - like we do in https://github.com/koreader/koreader/blob/c8b3942bf45ef70b4018b2d1d8272c19d8c0eb57/frontend/apps/reader/modules/readerrolling.lua#L226-L263

poire-z:
Just mentionning something I just saw while trying to implement support for CSS min-width/max-height:

In lvtextfm.cpp, there is a resizeImage() function that is supposed to reduce if needed the image size to fit in the paragraph width/page height, but also to apply some of CoolReader options, the Image scaling options.
We don't have these options in KOReader, and never needed to care about them.
But I see that with this PR, you picked modifications from https://github.com/koreader/crengine/commit/9214bfc59#diff-e6da683d521a05f10d1a627822b9d41bbd6646c190f0ec869d779ca27a3c19be (https://github.com/koreader/crengine/pull/35, support for CSS width/height on images, from 2016, before I was involved), that with commenting all the `// maxScale = ` made it always `maxScale=1`.
And the fact that maxScale=1 makes all CoolReader Image scaling options actually do nothing (well, that's how it looks to me, haven't tried it).

Looks like nobody noticed it since this was merged :) and may be it's just fine.
So, not thinking much about it - but if you were willing to remake it work (we could probably in KR force the image scaling options to do as with maxScale=1) - because how these scaling options should behave vs. CSS styles' width/min-width/max-width is some thinking I don't want to get into :)

poire-z:
For info, I restored the maxScale stuff in a commit part of https://github.com/koreader/crengine/pull/409 : 
```
resizeImage(): restore original scaling options code

It was hardcoded'ly removed as part of 9214bfc when
implementing support for CSS width & height for images,
so they don't get in the way.
Removed the hardcoding so the original CoolReader
settings can work again.
We can still disable that from frontend code.
```

Issue #137, opened 2020-05-10
3.2.39 broken table view and some others


ssten:
After updating to 3.2.39 I've find that coolreader begins incorrectly display tables, missing empty lines and change behaviour in display documet's sections.

Previous (expected) view
![3 2 2 1-r2](https://user-images.githubusercontent.com/2209401/81501718-77e54600-92e2-11ea-8fc1-d7372a0bf5a4.png)

3.2.39 view:
![3 2 39](https://user-images.githubusercontent.com/2209401/81501719-79167300-92e2-11ea-9180-05f2399f9fcb.png)

poire-z:
> After updating to 3.2.39 I've find that coolreader begins incorrectly display tables, missing empty lines and change behaviour in display documet's sections.

I don't see the missing lines, and dunno about document sections. Is that a FB2 document?

For the changes in tables rendering, I don't think it's incorrect in an EPUB context (but may be it is in the FB2 context): it's some change introduced in KOReader (and picked by CR 3.2.29) to display tables as the CSS specs say and as browsers do (see https://github.com/koreader/crengine/pull/243).
If you wish them to take all the screen width, add into your fb2.css or epub.css:
`table { width: 100% !important; }`
And if you want borders:
`table, tcaption, tr, th, td { border: black solid 1px; border-collapse: collapse; }`

If that's expected for FB2 (where publishers may not style tables), well, it might then be added in the shipped fb2.css (but not in epub.css, as EPUB publishers may expect "shrink to fit" and "no border" to be the default, and tweak that only when needed).

ssten:
It is the same fb2 file, generated from first example of OOoFBTools.
ExportToFB21_example-1.fb2

virxkane:
@ssten Is it still relevant, or can it be closed?

ssten:
3.2.57, table display is OK.

Issue #140, opened 2020-05-24
3.2.39 (fb2) hidden text formatted with built-in styles


ssten:
One of historical advantages of CoolReader was the support of styling, including built-in styles for spacing text formatting.

In 3.2.39 this feature is broken.

I.e. file:

```
<?xml version="1.0" encoding="utf-8"?>
<FictionBook xmlns="http://www.gribuser.ru/xml/fictionbook/2.0" xmlns:l="http://www.w3.org/1999/xlink">
<stylesheet type="text/css">
  style[name="razryadka"] {
  letter-spacing: 0.4em;
  }
</stylesheet>
  <description>
…
<p>One, <style name="razryadka">two</style>, three.</p>
…
```

Displays only:

> One, , three.

poire-z:
Didn't know that `<style>` was a proper FB2 element.
In EPUB, it's mostly used in the `<HEAD>` element, and I had hardcoded it to be hidden by default (because when disabling stylesheets, it would otherwise show):
https://github.com/buggins/coolreader/blob/26014a17806d91615d48214d19d29579c8992d42/crengine/include/fb2def.h#L48

You can have it working again I think by adding in fb2.css:
```css
style { display: inline; }
```

virxkane:
Fixed.
Screenshot from current master branch:
![Screenshot_20210820_092937](https://user-images.githubusercontent.com/36960933/130184263-ce1e555c-6f80-4aa6-b625-60dda259125d.png)

PR #142, opened 2020-07-31
Koreader merge post

virxkane:
koreader-merge-post

1. Merge latest koreader/crengine features. Some of them listed below:
   * Lot of bugs/SEGFAULTs fixes.
   * CSS/Text: adds "-cr-hint: strut-confined"
   * Tables cells: handle vertical-align: baseline
   * Hardcoded elements list: add <legend>
   * Tables: complete incomplete tables
   * CSS font-family parsing: set usual default of sans-serif
   * FB2 footnotes: fix spurious newline ("display: run-in")
   * CSS: adds support for font-variant
   * Add support for 
   * Use libunibreak for line breaking
   * Adds TextLangMan for text typography by language
   * Text: fix read/write outside array bounds
   * Font: fix wrong usage of Harfbuzz y_offset (#340)
   * Text: tunable use of letter spacing for justification
   * GIF decoding: avoid uneeded abort on LZW table overflow, avoid crash on some images
   * CSS: support for pseudo elements ::before & ::after
   * Support ePub 3 series & cover metadata (#346)
   * Adds support for "display: ruby" and <ruby> elements
   * CSS: adds support for "background-size" property
   * CSS: support background:url("data:image/png;base64,...)
   * CSS: adds text-align-last: -cr-right-if-not-first private keywords
   * Fix possible crash on book with embedded fonts load
   * Hanging punctuation: new implementation (#355)

   In fact, this list should be the largest, but in order not to overload the text, I had to shorten it, the full list of changes can be viewed with the git log command.

2. Adaptation for new koreader/crengine features.
   * Added option to settings dialog for selection between various presets of
    the rendering flags: engine option PROP_RENDER_BLOCK_RENDERING_FLAGS.
   * Added into settings dialog option for DOM level selection.
   * Added thirdparty library fribidi to support RTL.
   * Added thirdparty library libunibreak.
   * Added to database for each documents: 'requested DOM version', 'block rendering flags'.
   * Added option into OptionsDialog for 'requested DOM version' - can be changed for each file separate. Options 'requested DOM version' together with 'block rendering flags' allow processing old book mark saved in DB in previous version of CoolReader.
   * Updated fb2.css, fb3.css, epub.css, html5.css to support new features & restore old.
   * Updated settings dialog for new TextLangMan properties...

   Adaptation not finished, some options added in future PRs.

3. Multiple SEGFAULT fixed
   * Null-pointer dereference in in renderListItemMarker()
   * Fixed invalid cache usage: if the document buffer size is too small, ldomTextStorageChunk::ensureUnpacked() can remove itself from the buffer/memory. And as result, abnormal program termination.
   * Sanitizer fixes: multiple memory errors have been fixed - heap-buffer-overflow, use-after-free,
 index out of range, etc.
        This fixes oldest big bug with invalid cache usage:
        https://github.com/koreader/crengine/pull/358#issuecomment-662677793
   * Fixed heap-buffer-overflow in crengine/src/lvtextfm.cpp: LVFormatter::copyText(int, int)
   * AddressSanitizer: heap-buffer-overflow fixes. Referring to the array element at index -1.
   * Fixed page margins for compatibility with new hanging punctuation introduced in 6c2922c9cf26767
7e2bfed1969093066b80fc18d.

4. Other changes:
   * Added to database for each documents: CRC32. CRC32 added for future finding file in DB by CRC32 and file size, but not only file name.
   * Android: fixes incorrect/obsolete value of field 'format' in table 'book'. After adding support for the 'fb3' and 'docx' formats in version 3.2.33, the 'format' field in the 'book' table becomes invalid because the enum DocumentFormat has been changed. So, after reading this field from the database, we must recheck the format by pathname.
   * Fixed integer overflow in cache size calculations, relevant for very large cache sizes.
   * Small improvements to increase performance in org.coolreader.crengine.DocView.getPositionPropsI
nternal() that is called few time at each page turn.
   * Fixed DecodeWtf8() function.
   * Support for multiple fallback fonts (firstly implemented by koreader/crengine, but my implementation writen from scratch).

Fixes issues #137, #140.
I think we can make a new release.


crengine/src/lvtinydom.cpp
```diff
@@ -7478,7 +7478,6 @@ ldomDocumentWriter::~ldomDocumentWriter()
         _document->getRootNode()->initNodeFont();
         //if ( !_document->validateDocument() )
         //    CRLog::error("*** document style validation failed!!!");
-        _document->updateRenderContext();
```
poire-z, line 6009: Can you tell me a bit more of why/when that causes an issue ?
Could it be something that affects CoolReader only, and not KOReader, because we might do things a bit differently ?
(Just asking, I haven't really looked at why that's done here - but I haven't noticed any issue, or if I did I fixed any elsewhere.)

virxkane: Try this file from issue #140. 
```fb2
<?xml version="1.0" encoding="utf-8"?>
<FictionBook xmlns="http://www.gribuser.ru/xml/fictionbook/2.0" xmlns:l="http://www.w3.org/1999/xlink">
<stylesheet type="text/css">
  style[name="razryadka"] {
  letter-spacing: 0.4em;
  color: red;
  }
</stylesheet>

<p>One, <style name="razryadka">two</style>, three.</p>

</FictionBook>
```
If word ```two``` not red, you affected.

virxkane: It doesn’t even look right logically. We just loaded the file, but haven't rendered yet, but we keep the style hashes as after rendering.

poire-z: OK, so I'm affected too :) 
But your fix cause a full re-rendering to always be done after loading it seems.
With 3 of my usual EPUB documents, I get this - and I don't think they should or need that re-rendering:
```
08/01/20-13:54:35 DEBUG CreDocument: loading document...
08/01/20-13:54:39 DEBUG CreDocument: loading done.
08/01/20-13:54:39 DEBUG CreDocument: rendering document...
CRE: styles re-init needed after load, re-rendering
08/01/20-13:54:46 DEBUG CreDocument: rendering done.
```
So, doubling the load time.

It seems I get no issue and see it red with this in some HTML file:
```html
<html>
<head>
<style>
span[name=toto] { color: red; }
</style>
</head>
<body>
<H1>Margin vs <span name=toto>padding</span> vs borders</H1>
```

It fails also with just `<stylesheet type="text/css"> b { color: blue; } </stylesheet>` on the first loading - changing font size to have a re-rendering make it work.

So, the issue might more probably be with the `<stylesheet>` specific handling in FB2 (I think I remember it was specific - and it might be some spaghettugly stuff...) - that may be I've just never updated to work better. In the worst case we could just force that re-rendering when we meet such `<stylesheet>`. I'll have a look later.

poire-z: > We just loaded the file, but haven't rendered yet, but we keep the style hashes as after rendering.

It's an original crengine optimisation:

On first loading, when parsing the HTML, we compute styles of a node as we meet/enter it (because we can: we know its style and the style of its parent for inheritance) - and we compute rendering method when leaving the node (as then, we know all the rendering methods of its children). That way, after `~ldomDocumentWriter()`, styles and rendering methods are set - and we can just go at rendering (lvrend.cpp).

On re-rendering, the XHTML is valid, but we have to re-initNodeStyles (recursively, top first) and re-initNodeRendMethod (recursively, deep first) - and then, do the rendering (lvrend.cpp).

This double behaviour has always been painful, and one needs to check changes work in both codepaths (as you see with that FB2 `<stylesheet>`, it doesnt work in the 1st path). So, painful, but a nice optimisation I never dared dropping :)

poire-z: > If word two not red, you affected.

If you have some older CoolReader version (before any pick of my stuff), can you check if this ever worked on first load?
(Just so I know if I broke it - or if it was initially overlooked)

poire-z: I guess I have broken it with https://github.com/koreader/crengine/commit/9cecd154b where I commented out in `ldomDocumentWriter::OnTagClose()` what seems needed with FB2:` if isStyleSheetTag => document->applyDocumentStyleSheet();`
`This is now dealt with in :OnTagBody(), just before creating this <stylesheet> tag` was fine with HTML/EPUB - but with FB2, there is no body and it has to be done on the closing of the <stylesheet> tag as the real content is its sibling - and not children.

poire-z: This seems the right way to handle it:
```diff
--- a/crengine/src/lvtinydom.cpp
+++ b/crengine/src/lvtinydom.cpp
@@ -7629,6 +7629,7 @@ void ldomDocumentWriter::OnTagClose( const lChar16 *, const lChar16 * tagname )
         }
     }
     */
+    bool isStyleSheetTag = tagname[0] == 's' && tagname[1] == 't' && _currNode && !lStr_cmp(tagname, "stylesheet");

     lUInt16 id = _document->getElementNameIndex(tagname);
     //lUInt16 nsid = (nsname && nsname[0]) ? _document->getNsNameIndex(nsname) : 0;
@@ -7643,11 +7644,25 @@ void ldomDocumentWriter::OnTagClose( const lChar16 *, const lChar16 * tagname )
         _parser->Stop();
     }

-    /* This is now dealt with in :OnTagBody(), just before creating this <stylesheet> tag
-    if ( isStyleSheetTag ) {
+    // For EPUB/HTML, this is now dealt with in :OnTagBody(), just before creating this <stylesheet> tag.
+    // But for FB2, where we have:
+    //   <FictionBook>
+    //     <stylesheet type="text/css">
+    //       some css
+    //     </stylesheet>
+    //     <p>...
+    //     other content
+    //   </FictionBook>
+    // we need to apply the <stylesheet> content we have just left, so it applies
+    // to the coming up content.
+    // We check the parent we have just pop'ed is a <FictionBook>.
+    // Caveat: any style set on the <FictionBook> element itself won't be applied now
+    // in this loading phase (as we have already set its style) - but it will apply
+    // on re-renderings.
+    if ( isStyleSheetTag && _currNode->getElement()->getNodeId() == el_FictionBook ) {
         //CRLog::trace("</stylesheet> found");
 #if BUILD_LITE!=1
-        if ( !_popStyleOnFinish ) {
+        if ( !_popStyleOnFinish && _document->getDocFlag(DOC_FLAG_ENABLE_INTERNAL_STYLES) ) {
             //CRLog::trace("saving current stylesheet before applying of document stylesheet");
             _document->getStyleSheet()->push();
             _popStyleOnFinish = true;
@@ -7655,7 +7670,6 @@ void ldomDocumentWriter::OnTagClose( const lChar16 *, const lChar16 * tagname )
         }
 #endif
     }
-    */

     //logfile << " !c!\n";
 }
```
with the caveat mentionned in the comment - not sure if that's critical.
Also, there was/is some `#if BUILD_LITE!=1` in there - in the other stuff I did in the referenced commit above, I think I didn't wrap anything in `#if BUILD_LITE!=1` - so it's a bit inconsistent.
I dunno if CoolReader is still built somewhere with `BUILD_LITE=1` - but if not, you could remove it there - but if yes, it might need a bit more wrapping in a few other places.

virxkane: Just found some free time. Did you just return half of what you commented out? :) Ok it works, but you puts too much checks for not null value of _currNode, this check is at the begin of function onTagClose(). As I remember, you don't like unnecessary checks :)
Thank you for patch.

poire-z: > Did you just return half of what you commented out? :) 

Yep, I wasn't much aware of the FB2 stuff at the time :)

> you puts too much checks for not null value of _currNode, this check is at the begin of function onTagClose(). 

Indeed, feel free to clean that up.

I prefer my `bool isStyleSheetTag = tagname[0] == 's' && tagname[1] == 't' && _currNode && !lStr_cmp(tagname, "stylesheet");` just because we might meet a lot of other tags that start with "s" like `<span>`, and checking for "t" avoids the more expensive `Str_cmp()` on all of them.

PR #144, opened 2020-08-03
Restored sentence selection in TTS that broken in 3.2.39 resulting in…

virxkane:
Restored sentence selection in TTS that broken in 3.2.39 resulting in a merge with koreader.
See https://github.com/koreader/crengine/pull/69#issuecomment-668175384
Before:
before
And after:
after


virxkane:
Also tested with inner tags, for example `<b>`
[1.fb2.zip](https://github.com/buggins/coolreader/files/5020382/1.fb2.zip)
with-b-tag

buggins:
cool! thank you!

PR #146, opened 2020-08-06
[Android] Use Java 8 language features.

Isira-Seneviratne:
* Enable [Java 8 support](https://developer.android.com/studio/write/java8-support).
* Make use of Java 8 language features such as lambda expressions and method references.

### DIFF
--- a/android/src/org/coolreader/crengine/Utils.java
+++ b/android/src/org/coolreader/crengine/Utils.java
@@ -89,13 +89,10 @@ public class Utils {
 			Log.i("cr3", "File " + oldPlace.getAbsolutePath() + " does not exist!");
 			return false;
 		}
-		FileOutputStream os = null;
-		FileInputStream is = null;
-		try {
+		try (FileOutputStream os = new FileOutputStream(newPlace);
+			FileInputStream is = new FileInputStream(oldPlace)) {
 			if (!newPlace.createNewFile())
 				return false; // cannot create file
-			os = new FileOutputStream(newPlace);
-			is = new FileInputStream(oldPlace);
 			copyStreamContent(os, is);
 			removeNewFile = false;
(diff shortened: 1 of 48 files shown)


virxkane:
And what does it give?
One more note, the onyx sdk did not need to be touched at all.

Isira-Seneviratne:
> And what does it give?

It makes the code much simpler.

> One more note, the onyx sdk did not need to be touched at all.

Okay, thanks for the info.

android/src/org/coolreader/crengine/Utils.java
```diff
@@ -89,32 +89,17 @@ private static boolean moveFile(File oldPlace, File newPlace, boolean removeOld)
 			Log.i("cr3", "File " + oldPlace.getAbsolutePath() + " does not exist!");
 			return false;
 		}
-		FileOutputStream os = null;
-		FileInputStream is = null;
-		try {
+		try (FileOutputStream os = new FileOutputStream(newPlace);
```
virxkane, line 92: @Isira-Seneviratne You have changed the order of function calls.
Before:
```java
if (!newPlace.createNewFile())
	return false; // cannot create file
os = new FileOutputStream(newPlace);
is = new FileInputStream(oldPlace);
```
After:
```java
FileOutputStream os = new FileOutputStream(newPlace);
FileInputStream is = new FileInputStream(oldPlace);
if (!newPlace.createNewFile())
	return false;
```
So, now this function will always return false.
https://developer.android.com/reference/java/io/FileOutputStream#FileOutputStream(java.io.File)
https://developer.android.com/reference/java/io/File#createNewFile()

Isira-Seneviratne: Okay, I'll fix it.

PR #149, opened 2020-08-06
[Android] Use the Java 8 Date/Time API.

Isira-Seneviratne:
* Enable [Java 8 API desugaring](https://developer.android.com/studio/write/java8-support#library-desugaring).
* Increase the minimum API level to 4 (the minimum supported by multidex).
* Use the Java 8 Date/Time API in `Utils`.

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -4,7 +4,7 @@ android {
     compileSdkVersion 28
     defaultConfig {
         applicationId "org.coolreader"
-        minSdkVersion 3
+        minSdkVersion 4 // This is the minimum supported by multidex
         targetSdkVersion 28
         // When new version released, version code must be incremented at least by 8
         // for compatibility with ABI versioning of split apk (see below).
@@ -17,6 +17,7 @@ android {
                 cppFlags ""
             }
         }
+        multiDexEnabled true
     }
     sourceSets {
         main {
@@ -118,10 +119,16 @@ android {
         // Sets Java compatibility to Java 8
         sourceCompatibility JavaVersion.VERSION_1_8
         targetCompatibility JavaVersion.VERSION_1_8
+
+        // Flag to enable support for the new language APIs
+        coreLibraryDesugaringEnabled true
     }
 }
 
 dependencies {
+    coreLibraryDesugaring 'com.android.tools:desugar_jdk_libs:1.0.10'
+    implementation 'com.android.support:multidex:1.0.3'
+
     implementation fileTree(include: ['*.jar'], dir: 'libs')
     testImplementation 'junit:junit:4.12'
     androidTestImplementation 'com.android.support.test:runner:1.0.2'
--- a/android/app/src/main/AndroidManifest.xml
+++ b/android/app/src/main/AndroidManifest.xml
@@ -26,13 +26,14 @@
     <uses-permission android:name="com.android.vending.BILLING" />
 
     <application
-        android:allowBackup="true"
-        android:allowClearUserData="true"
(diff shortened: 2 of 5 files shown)


virxkane:
All the same, tell me, why are you doing this, why are you promoting Java 8 features? You say the code will be simpler, a controversial statement, but okay.
Do you want to make any changes to this project (except for Java 8 features) and for this you bring the source code into a form that suits you?
Or just translate codebase to Java 8? What for?

virxkane:
I hope you are testing all the replaced code? On all supported versions of Android? Then we don't have to answer the questions, why is my favorite program not working anymore?

Isira-Seneviratne:
> All the same, tell me, why are you doing this, why are you promoting Java 8 features? You say the code will be simpler, a controversial statement, but okay.
> Do you want to make any changes to this project (except for Java 8 features) and for this you bring the source code into a form that suits you?
> Or just translate codebase to Java 8? What for?
> 
> 

I thought the new features would be helpful, and they aren't restricted by API level anymore: https://jakewharton.com/androids-java-8-support/

virxkane:
So you translate the sources to Java 8 for some abstract good?

Isira-Seneviratne:
> So you translate the sources to Java 8 for some abstract good?

Yeah, it doesn't affect the functionality.

Isira-Seneviratne:
> > So you translate the sources to Java 8 for some abstract good?
> 
> Yeah, the new language features don't affect the functionality.

virxkane:
> Yeah, it doesn't affect the functionality.

Can you test your changes on Android 4.0, Android 2.3? Yes, of course, it seems like it should work, but have you tested?

virxkane:
Will there be no performance losses?
In addition to the fact that Java 8 features should work on older versions of Android, you could accidentally make bugs when translating to Java 8, which affected most of the code, so this cannot without extensive testing.
That is why I ask, did you test at least the main functionality of the program?

Isira-Seneviratne:
Yeah, I ran the app and didn't encounter any issues.

Isira-Seneviratne:
> Will there be no performance losses?
> In addition to the fact that Java 8 features should work on older versions of Android, you could accidentally make bugs when translating to Java 8, which affected most of the code, so this cannot without extensive testing.
> That is why I ask, did you test at least the main functionality of the program?

I made the changes with the help of Android Studio. It automatically converted interface declarations to lambdas.

virxkane:
> Yeah, I ran the app and didn't encounter any issues.

Android 2.2, 2.3, 4.0.4, 4.2, etc ???

Isira-Seneviratne:
> > Yeah, I ran the app and didn't encounter any issues.
> 
> Android 2.2, 2.3, 4.0.4, 4.2, etc ???

Testing the language features on all API levels isn't necessary, because Gradle converts the code to run on all API levels: https://jakewharton.com/androids-java-8-support/#source-transformation

virxkane:
Yes, of course, but is there any disadvantages in this desugaring, for example, slowing down?

Isira-Seneviratne:
I didn't encounter any performance issues or anything like that.

Isira-Seneviratne:
I encountered issues with using an AVD running Android Donut, but it seems the solution is to install an older version of the Android SDK: https://stackoverflow.com/questions/60187172/emulate-android-1-6-donut-with-arm-abi

If this PR works, I'll reopen it.

Isira-Seneviratne:
Unfortunately, I couldn't get the emulator to run Android versions older than ICS.

PR #160, opened 2020-08-30
Fixed footnotes rendering in legacy rendering mode that is broken in …

virxkane:
Fixed footnotes rendering in legacy rendering mode that is broken in afc16d0838767a9e16f29ce5e4bb98d63e03a7ef.
See https://github.com/buggins/coolreader/pull/157#issuecomment-683417317

### DIFF
--- a/crengine/src/lvrend.cpp
+++ b/crengine/src/lvrend.cpp
@@ -2502,6 +2502,10 @@ void renderFinalBlock( ldomNode * enode, LFormattedText * txform, RenderRectAcce
         int direction = RENDER_RECT_PTR_GET_DIRECTION(fmt);
         bool is_rtl = direction == REND_DIRECTION_RTL;
 
+        ldomNode * parent = enode->getParentNode(); // Needed for various checks below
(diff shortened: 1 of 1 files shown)


crengine/src/lvrend.cpp
```diff
@@ -2511,7 +2515,24 @@ void renderFinalBlock( ldomNode * enode, LFormattedText * txform, RenderRectAcce
         // In legacy rendering mode, we should get the same text formatting flags
         // as in CoolReader 3.2.38 and earlier, i.e. set is_block to true for
         // any block elements.
-        bool is_block = BLOCK_RENDERING_G(ENHANCED) ? rm == erm_final : style->display >= css_d_block;
+        bool is_block = rm == erm_final;
+        if (!BLOCK_RENDERING_G(ENHANCED) && !is_block) {
```
poire-z, line 2519: May be now that you made it a `if`, move your 3 comment lines above ("// In legacy rendering mode...") inside it - so one can more easily skip reading comments regarding legacy hacks.

PR #164, opened 2020-09-05
ODF (odt) format support.

pkb:
The format support level is about the same as docx's.  As usual I tested using Qt build. For android some adaptation will be required.


cr3qt/data/docx.css
```diff
@@ -50,19 +50,23 @@ ol { display: block; list-style-type: decimal; margin-left: 1em }
 .h4 { font-size: 120% }
 .h5 { font-size: 110% }
 
-table { font-size: 80% }
+table { font-size: 80%; width: 100%; }
 td, th { text-indent: 0px; padding: 3px }
-th {  font-weight: bold; text-align: center; background-color: #DDD  }
+/* table borders */
+table, th, td { border-width: 2px; border-style: ridge; border-color: #888; border-collapse: collapse; }
```
Frenzie, line 56: This is in order to be consistent with tables in the EPUB style? I'd normally expect a simple solid border in DOCX.

crengine/include/crsetup.h
```diff
@@ -131,7 +131,9 @@
 #else
 #define USE_FREETYPE                         1
 #define USE_HARFBUZZ                         1
+#ifndef _MSC_VER
```
Frenzie, line 134: This kind of preprocessor stuff strikes me as more a matter for CMakeLists? Put another way, surely MSC isn't incompatible with FriBidi, just a little harder to setup?

Frenzie (commented):
Very neat! ^_^

cr3qt/data/docx.css
pkb: I just copied it from fb2.css without any further thought :) The proper way would be to read border properties for the docs  but I'm not going to do it at least in a foreseeable future. So I'll leave it as is for now and let others decide how should it be.

crengine/include/crsetup.h
pkb: Yes, I agree but I don't wan to mess with this CMake stuff. We use it as a static library and it doesn't work well with MSVC. If someone will be interested to build CR with fribidi he will have to fix it properly.

virxkane:
The text style is not always applied correctly.
File sample3.odt:
[sample3.zip](https://github.com/buggins/coolreader/files/5187481/sample3.zip)

Comparision LibreOffice 6.4 and CoolReader current:
samplr3-compare

pkb:
Fixed it in c00d9455, if you will not find any other bugs I will create PR. Note italic style will still be applied to ЙЦУКЕН, it comes from subtitle default css setting.

virxkane:
@pkb, Ok, thank you.

PR #171, opened 2020-09-26
Refactoring some globals variable in lvrend.h

virxkane:
Refactoring some globals variable in lvrend.h
* The "gHangingPunctuationEnabled" global variable has been transformed to the "_hangingPunctuationEnabled" property in the "tinyNodeCollection" class.
* The "gRenderBlockRenderingFlags" global variable has been transformed to the "_renderBlockRenderingFlags" property in the "tinyNodeCollection" class.
* The "gDOMVersionRequested" global variable has been transformed to the "_DOMVersionRequested" property in the "tinyNodeCollection" class.
* The "gInterlineScaleFactor" global variable has been transformed to the "_interlineScaleFactor" property in the "tinyNodeCollection" class.

Desktop/Qt build use at least two document instances. The second document is used as a preview in the settings dialog. So, if we use these globals (more related to gHangingPunctuationEnabled) and apply changes to this document when using the preview, then we prevent the changes from being applied to the main document, since the value of these global variables is checked when the changes are applied.
Also, in the Android build in the future, a preview may appear in the settings dialog :)

### DIFF
--- a/crengine/include/lvrend.h
+++ b/crengine/include/lvrend.h
@@ -126,12 +126,12 @@ lUInt32 styleToTextFmtFlags( bool is_block, const css_style_ref_t & style, lUInt
 /// renders block as single text formatter object
 void renderFinalBlock( ldomNode * node, LFormattedText * txform, RenderRectAccessor * fmt, lUInt32 & flags,
                        int indent, int line_h, TextLangCfg * lang_cfg=NULL, int valign_dy=0, bool * is_link_start=NULL );
-/// renders block which contains subblocks (with gRenderBlockRenderingFlags as flags)
+/// renders block which contains subblocks (with enode document's getRenderBlockRenderingFlags() as flags)
 int renderBlockElement( LVRendPageContext & context, ldomNode * enode, int x, int y, int width,
         int usable_left_overflow=0, int usable_right_overflow=0, int direction=REND_DIRECTION_UNSET, int * baseline=NULL );
 /// renders block which contains subblocks
 int renderBlockElement( LVRendPageContext & context, ldomNode * enode, int x, int y, int width,
-        int usable_left_overflow, int usable_right_overflow, int direction, int * baseline, int rend_flags );
+        int usable_left_overflow, int usable_right_overflow, int direction, int * baseline, lUInt32 rend_flags );
 /// renders table element
 int renderTable( LVRendPageContext & context, ldomNode * element, int x, int y, int width,
                  bool shrink_to_fit, int & fitted_width, int direction=REND_DIRECTION_UNSET,
@@ -179,8 +179,6 @@ extern int gRootFontSize;
 #define INTERLINE_SCALE_FACTOR_SHIFT 10
 extern int gInterlineScaleFactor;
 
-extern int gRenderBlockRenderingFlags;
-
 // Enhanced rendering flags
 #define BLOCK_RENDERING_ENHANCED                           0x00000001
 #define BLOCK_RENDERING_ALLOW_PAGE_BREAK_WHEN_NO_CONTENT   0x00000002 // Allow consecutive page breaks when only separated
@@ -221,7 +219,6 @@ extern int gRenderBlockRenderingFlags;
 // Enable everything
 #define BLOCK_RENDERING_FULL_FEATURED                      0x7FFFFFFF
 
-#define BLOCK_RENDERING_G(f) ( gRenderBlockRenderingFlags & BLOCK_RENDERING_##f )
 #define BLOCK_RENDERING(v, f) ( v & BLOCK_RENDERING_##f )
 
 // rendering flags presets
@@ -247,7 +244,5 @@ extern int gRenderBlockRenderingFlags;
                                            BLOCK_RENDERING_BOX_INLINE_BLOCKS )
 #define BLOCK_RENDERING_FLAGS_WEB        BLOCK_RENDERING_FULL_FEATURED
 #define BLOCK_RENDERING_FLAGS_DEFAULT    BLOCK_RENDERING_FLAGS_WEB
-
-int validateBlockRenderingFlags( int f );
 
 #endif
--- a/crengine/include/lvrend.h
+++ b/crengine/include/lvrend.h
@@ -177,7 +177,6 @@ extern int gRootFontSize;
 
 #define INTERLINE_SCALE_FACTOR_NO_SCALE 1024
 #define INTERLINE_SCALE_FACTOR_SHIFT 10
-extern int gInterlineScaleFactor;
 
 // Enhanced rendering flags
 #define BLOCK_RENDERING_ENHANCED                           0x00000001
--- a/crengine/include/lvrend.h
+++ b/crengine/include/lvrend.h
@@ -175,8 +175,6 @@ extern int gRenderDPI;
 extern bool gRenderScaleFontWithDPI;
 extern int gRootFontSize;
 
-extern bool gHangingPunctuationEnabled;
-
 #define INTERLINE_SCALE_FACTOR_NO_SCALE 1024
 #define INTERLINE_SCALE_FACTOR_SHIFT 10
 extern int gInterlineScaleFactor;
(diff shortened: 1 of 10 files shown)


crengine/src/lvstsheet.cpp
```diff
@@ -4435,7 +4435,7 @@ bool LVStyleSheet::parse( const char * str, bool higher_importance, lString16 co
             }
             // parse declaration
             LVCssDeclRef decl( new LVCssDeclaration );
-            if ( !decl->parse( str, higher_importance, _doc, codeBase ) )
+            if ( !decl->parse( str, (_doc != NULL) ? _doc->getDOMVersionRequested() : 0, higher_importance, _doc, codeBase ) )
```
poire-z, line 4438: May be set `lUInt32 domVersionRequested = _doc != NULL) ? _doc->getDOMVersionRequested() : 0;` at top of the function to not have to re-compute it it as each new declaration.

crengine/src/lvtinydom.cpp
```diff
@@ -14112,40 +14141,37 @@ ldomDocumentWriterFilter::ldomDocumentWriterFilter(ldomDocument * document, bool
 , _curFosteredNode(NULL)
 , _lastP(NULL)
 {
-    if (gDOMVersionRequested >= 20200824) {
+    if (_document->getDOMVersionRequested() < 20200824) {
```
poire-z, line 14153: Why change the comparison order?  You then have to indent what follows, which cause some diff (and I wonder if you changed something - I guess not - and it will pollute git blame).

crengine/src/lvtinydom.cpp
```diff
@@ -14112,40 +14141,37 @@ ldomDocumentWriterFilter::ldomDocumentWriterFilter(ldomDocument * document, bool
...
 ldomDocumentWriterFilter::~ldomDocumentWriterFilter()
 {
-
-    if (gDOMVersionRequested >= 20200824) {
-        return;
-    }
-    for ( int i=0; i<MAX_ELEMENT_TYPE_ID; i++ ) {
-        if ( _rules[i] )
-            delete[] _rules[i];
+    if (_document->getDOMVersionRequested() < 20200824) {
```
poire-z, line 14179: Same comment as above.

crengine/include/lvrend.h
```diff
@@ -221,7 +219,6 @@ extern int gRenderBlockRenderingFlags;
 // Enable everything
 #define BLOCK_RENDERING_FULL_FEATURED                      0x7FFFFFFF
 
-#define BLOCK_RENDERING_G(f) ( gRenderBlockRenderingFlags & BLOCK_RENDERING_##f )
 #define BLOCK_RENDERING(v, f) ( v & BLOCK_RENDERING_##f )
```
poire-z, line 221: May be we could have a few other macros, to reduce the length of the line that do this:
```
BLOCK_RENDERING(getDocument()->getRenderBlockRenderingFlags(), BOX_INLINE_BLOCKS)
BLOCK_RENDERING(enode->getDocument()->getRenderBlockRenderingFlags(), BOX_INLINE_BLOCKS)
could become:
BLOCK_RENDERING_N(this, BOX_INLINE_BLOCKS)
BLOCK_RENDERING_N(enode, BOX_INLINE_BLOCKS)
with 
#define BLOCK_RENDERING_N(n, f) ( n->getDocument()->getRenderBlockRenderingFlags() & BLOCK_RENDERING_##f )

BLOCK_RENDERING(doc->getRenderBlockRenderingFlags(), BOX_INLINE_BLOCKS)) 
could become:
BLOCK_RENDERING_D(this, BOX_INLINE_BLOCKS)
with 
#define BLOCK_RENDERING_D(d, f) ( d->getRenderBlockRenderingFlags() & BLOCK_RENDERING_##f )
```

crengine/src/lvrend.cpp
```diff
@@ -3805,7 +3793,7 @@ int pagebreakhelper(ldomNode *enode,int width)
 
 // Prototypes of the 2 alternative block rendering recursive functions
 int  renderBlockElementLegacy(LVRendPageContext & context, ldomNode * enode, int x, int y, int width , int usable_right_overflow);
-void renderBlockElementEnhanced( FlowState * flow, ldomNode * enode, int x, int width, int flags );
+void renderBlockElementEnhanced(FlowState * flow, ldomNode * enode, int x, int width, lUInt32 flags );
```
poire-z, line 3795: My space ! My balance ! :)

poire-z:
I'm fine with the principle and the changes.
No time to look at the implications with KOReader, but not much (we use the setProp stuff, except for the gDomVersion where we do a bit more checks).

(I guess the multiple calls to the new functions (instead of just accessing a global variable), as they are defined in *.h files, will be inlined and won't cause much performance overhead?)

Maybe some other globals need to be looked at too:
```c
in crengine/include/lvrend.h
#define BASE_CSS_DPI 96 // at 96 dpi, 1 css px = 1 screen px
#define DEF_RENDER_DPI 96
#define DEF_RENDER_SCALE_FONT_WITH_DPI 0

extern int gRenderDPI; // we do change that from KOReader - not (yet :) on CoolReader
extern bool gRenderScaleFontWithDPI; // added for possibility, but we don't change it from KOReader
extern int gRootFontSize; // I think that it just for storing, but it changes when one changes font size
```

crengine/src/lvtinydom.cpp
virxkane: I don't think about 'git blame'. I think about code readability. Do you propose to revert this?

crengine/src/lvstsheet.cpp
virxkane: Ok.

crengine/include/lvrend.h
virxkane: To be honest, I would give them up altogether, they make the code difficult to read.

crengine/src/lvrend.cpp
virxkane: I don't know why this happened, maybe it's an IDE, they are too smart now :)

virxkane:
@poire-z 
> Maybe some other globals need to be looked at too:

I tried to refactor gRenderDPI a while ago, and maybe forgot the details a little.
I found it too complicated, and as far as I remember, this variable is deeply involved in the code associated with the style sheet where the document instance is not available.

poire-z:
> this variable is deeply involved in the code associated with the style sheet where the document instance is not available.

Right, that's true. So ok, best to let this one be a global variable and keep things simple.

crengine/include/lvrend.h
poire-z: Give up these macros ?! They help readability (a lot!) for me :)

crengine/src/lvtinydom.cpp
poire-z: >  I can revert this.

Yes please, thanks for git blame :)

virxkane:
> Do this preview needs a full-rerendering ?

Yes, full rerendering.

> Or there's some trick to prevent that ? :)

If I understood the question correctly, then 
https://github.com/koreader/crengine/blob/a00d1b8c26f4a621d805575a6a4cf10d81c8e7c7/crengine/src/lvdocview.cpp#L6517

virxkane:
> QT

QT - Quick Time, but Qt - it's framework by qt-project :)

virxkane:
> Do this preview needs a full-rerendering ? Or there's some trick to prevent that ? :)

If you mean preview in main document - NO, preview (with small text fragment) in small view in second document.

crengine/src/lvdocview.cpp
```diff
@@ -4645,6 +4645,13 @@ void LVDocView::createEmptyDocument() {
     m_doc->setMinSpaceCondensingPercent(m_props->getIntDef(PROP_FORMAT_MIN_SPACE_CONDENSING_PERCENT, DEF_MIN_SPACE_CONDENSING_PERCENT));
     m_doc->setUnusedSpaceThresholdPercent(m_props->getIntDef(PROP_FORMAT_UNUSED_SPACE_THRESHOLD_PERCENT, DEF_UNUSED_SPACE_THRESHOLD_PERCENT));
     m_doc->setMaxAddedLetterSpacingPercent(m_props->getIntDef(PROP_FORMAT_MAX_ADDED_LETTER_SPACING_PERCENT, DEF_MAX_ADDED_LETTER_SPACING_PERCENT));
+    m_doc->setHangingPunctiationEnabled(m_props->getBoolDef(PROP_FLOATING_PUNCTUATION, true));
```
poire-z, line 4648: Dunno if it was a typo or a decision, or if it even matters (is it always reset by frontend ? it would be the case with KOReader), but you set floating punctuation `true` (before, we had `bool gHangingPunctuationEnabled = false`), so it's enabled by default?

virxkane: I don't think that is important. Change it if you think it's necessary . Later I sync the sources.

Issue #181, opened 2020-10-18
Please Update the Windows Version its been 5 years! 


shockergit:
Please Update the Windows Version its been 5 years just give one update or do a YouTube video on how to make an exe file from source. Windows dont have any good ebook readers. even fb reader is discontinued for 10 years

ysrs:
how to build a windows version?

PR #183, opened 2020-10-19
String refactoring

virxkane:
Eliminated years of confusion with the lString16 class.
lChar16 & lString16 has been replaced by lChar32 & lString32, since in fact before lString16 was a string with 32-bit characters (wchar_t) in Linux and Android, and only in Windows - 16 bits.
This requires c++ standart "c++0x".
Then lString16 class added again, but only to be able to call WINAPI functions such as CreateFileW, DeleteFileW.
Now lString16 has 16-bit characters and lString32 has 32-bit characters on **all** platforms.

Previously, to compile for Windows, we had to use a patch like this:
```diff
diff --git a/crengine/src/lvxml.cpp b/crengine/src/lvxml.cpp
index a8f49ef5c..33901dd47 100644
--- a/crengine/src/lvxml.cpp
+++ b/crengine/src/lvxml.cpp
@@ -3163,8 +3163,13 @@ bool LVXMLParser::Parse()
 
 typedef struct  {
     const wchar_t * name;
+#ifdef _WIN32
+    char32_t code;
+    char32_t code2;
+#else
     wchar_t code;
     wchar_t code2;
+#endif
 } ent_def_t;
 
 // From https://html.spec.whatwg.org/multipage/named-characters.html#named-character-references 
```
And we could not open files whose names contain unicode characters whose codes are greater than 0xFFFF, for example emoji, nor could we see the contents of files that contain these characters. All this has now been fixed.
Using this test file we can check this: [test-emoji.zip](https://github.com/buggins/coolreader/files/5403603/test-emoji.zip).

Before this PR on Windows but with patch above:
File "4-byte-unicode-name_😀😁😂😃😄😅😆😇😈😉😊😋.txt" can't be opened:
Clipboard01
File "sample3.txt" opened, but incorrectly:
Clipboard02

And screenshots with this PR:
File "4-byte-unicode-name_😀😁😂😃😄😅😆😇😈😉😊😋.txt":
Clipboard03
File "sample3.txt":
Clipboard04

The files "4-byte-unicode-name_😀😁😂😃😄😅😆😇😈😉😊😋.txt" and "sample3.txt" are identical, they contain the following lines:
<pre>
File encoding is UTF8.
Some emoji: 😀😁😂😃😄😅😆😇😈😉😊😋
</pre>

The font used for rendering was "Noto Emoji" from https://github.com/googlefonts/noto-emoji as a fallback font.

This PR is related to issue #181 as it fixes compilation errors and flaws in build for Windows.

P.S. Don't use in new code class lString16, it's only for Windows support! Use only lString32 for unicode support.
P.P.S. The code in "crgui/src" hasn't been updated, I don't have the appropriate SDKs to make this work.


virxkane:
@pkb Can you try compiling CoolReader with this PR in Visual Studio?

PR #185, opened 2020-10-22
Added support for the '@charset' rule to the css-rules parsing code, …

virxkane:
Added support for the [`@charset`](https://developer.mozilla.org/en-US/docs/Web/CSS/@charset) rule to the css-rules parsing code, just so that no parsing error is generated, but then this rule is ignored.

This fixes the incorrect rendering of some epub/html files containing css files with such rules, for example, http://www.gutenberg.org/ebooks/11, or simplified test file [test2.zip](https://github.com/buggins/coolreader/files/5425090/test2.zip).

The crux of the problem is that if a '`@charset`' rule exists, the rule following it is completely ignored as if there was a parsing error.

Before:
before

In Mozilla Firefox:
firefox-s

After:
after

### DIFF
--- a/crengine/src/lvstsheet.cpp
+++ b/crengine/src/lvstsheet.cpp
@@ -4304,6 +4304,7 @@ LVStyleSheet::LVStyleSheet( LVStyleSheet & sheet )
 {
     set( sheet._selectors );
     _selector_count = sheet._selector_count;
+    _charset = sheet._charset;
 }
 
 void LVStyleSheet::apply( const ldomNode * node, css_style_rec_t * style )
@@ -4415,6 +4416,9 @@ bool LVStyleSheet::parse( const char * str, bool higher_importance, lString32 co
         bool err = false;
         for (;*str;)
         {
+            // parse charset, and... ignored it
+            // just to avoid generating a parse error
+            parseCharsetRule(str);
             // parse selector(s)
             // Have selector count number make the initial value
             // of _specificity, so order of selectors is preserved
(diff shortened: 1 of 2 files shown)


crengine/src/lvstsheet.cpp
```diff
@@ -4499,6 +4503,35 @@ bool LVStyleSheet::parse( const char * str, bool higher_importance, lString32 co
...
+{
+    // Parse rule '@charset' according https://developer.mozilla.org/en-US/docs/Web/CSS/@charset
+    if (!str || !*str)
+        return false;
+    skip_spaces( str );
+    if ( *str == '@' ) {
+        str++;
+        char word[64];
+        if ( parse_ident( str, word ) ) {
+            lString32 keyword(word);
```
poire-z, line 4516: (Really minor remark: keyword and _charset could just be lString8 - as we parse `char *` and we don't need any of these to be anything else.)
Looks fine to me.

virxkane: > Really minor remark: keyword and _charset could just be lString8 - as we parse char * and we don't need any of these to be anything else.

Ok, but lString8 don't have lowercase() method, but I don't know if it is possible to use capital letters in the 'charset' keyword.

poire-z: > but lString8 don't have lowercase() method

Oh, ok, I thought it had it - but indeed, no lowercase() :/. Well, then, may be use lString32 for the check, better be safe, everything is possible in the wild :)
(As for the charset itself, it's supposed to be ASCII, so I guess it's ok to be lString8 and we'll get raw chars in it.)

virxkane: Fixed, ignoring case :)

PR #187, opened 2020-10-22
Switch to Android target API 29 (Android 10).

virxkane:
Switch to Android target API 29 (Android 10).


virxkane:
Unstable, closed...

PR #200, opened 2020-11-22
Don't render 'WORD JOINER' character if no such glyph

virxkane:
* Fix index out of range in lString32::pos(lChar32) introduced in cc51fc6ff1336bc87ebaf146ee04b18207331d4a
* Do not render the Unicode character 'WORD JOINER' (U + 2060) as a replacement character if the font does not have a corresponding glyph for it in SHAPING_MODE_HARFBUZZ_LIGHT and SHAPING_MODE_FREETYPE shaping modes. Otherwise, it results in a question mark appearing in footnote number.
This symbol is embedded in the text after the d2369334b610f75504bc80cc0697cc19314710f0 commit.
See https://github.com/koreader/crengine/pull/385, https://github.com/koreader/koreader/issues/6718#issuecomment-699600019
Screenshot_20201122_163648-m

### DIFF
--- a/crengine/include/lvstring.h
+++ b/crengine/include/lvstring.h
@@ -26,6 +26,7 @@
 /// Unicode spaces
 #define UNICODE_NO_BREAK_SPACE            0x00A0
 #define UNICODE_ZERO_WIDTH_NO_BREAK_SPACE 0xfeff
+#define UNICODE_WORD_JOINER      0x2060
 // All chars from U+2000 to U+200B allow wrap after, except U+2007
 #define UNICODE_EN_QUAD          0x2000
 #define UNICODE_FIGURE_SPACE     0x2007
--- a/crengine/src/lvstring.cpp
+++ b/crengine/src/lvstring.cpp
@@ -2825,7 +2825,7 @@ int lString8::pos(const lString8 & subStr, int startPos) const
 }
 
 int lString32::pos(lChar32 ch) const {
-    for (int i = 0; i <= length(); i++)
+    for (int i = 0; i < length(); i++)
     {
         if (pchunk->buf32[i] == ch)
         {
@@ -2839,7 +2839,7 @@ int lString32::pos(lChar32 ch, int start) const
 {
     if (length() - start < 1)
         return -1;
-    for (int i = start; i <= length(); i++)
+    for (int i = start; i < length(); i++)
     {
         if (pchunk->buf32[i] == ch)
         {
--- a/crengine/src/private/lvfreetypeface.cpp
+++ b/crengine/src/private/lvfreetypeface.cpp
@@ -69,7 +69,7 @@ extern lString8 familyName(FT_Face face);
 
 inline int myabs(int n) { return n < 0 ? -n : n; }
 
-static lChar32 getReplacementChar(lUInt32 code) {
+static lChar32 getReplacementChar(lUInt32 code, bool * can_be_ignored = NULL) {
     switch (code) {
         case UNICODE_SOFT_HYPHEN_CODE:
             return '-';
@@ -80,8 +80,12 @@ static lChar32 getReplacementChar(lUInt32 code) {
         case UNICODE_NO_BREAK_SPACE:
             return ' ';
         case UNICODE_WORD_JOINER:
+            if (can_be_ignored)
+                *can_be_ignored = true;
             return UNICODE_ZERO_WIDTH_SPACE;
         case UNICODE_ZERO_WIDTH_SPACE:
+            if (can_be_ignored)
+                *can_be_ignored = true;
             // If the font lacks a zero-width breaking space glyph (like
             // some Kindle built-ins) substitute a different zero-width
             // character instead of one with width.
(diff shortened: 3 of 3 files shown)


virxkane:
@poire-z About 'WORD JOINER' do you think this is the right fix?

poire-z:
getReplacementChar() looks like probably the best place to handle that.
Having it speficied by name again in getCharIndex() feels a bit less nice.

May be we could have something like:
`static lChar32 getReplacementChar(lUInt32 code, bool * ignore_if_not_found=NULL)`
  where we could set this boolean (if not NULL) to true when WORD JOINER or any other kind of these things we could meet (I guess there are many more :) 
of have getReplacementChar returns 0xFFFFFFFF instead of 0 when it's fine to ignore it.

poire-z:
Unless it's ok to just flag it in here:
https://github.com/buggins/coolreader/blob/b01ce22844524e379b94d2ef176353ed0bf94d17/crengine/src/lvtextfm.cpp#L1366-L1380
This was so long ago - I'm not sure which is the best :)

When you ignore it in the drawing with this PR as it is, do you still some blank (taken by the "?" which is not drawn) ?

virxkane:
> May be we could have something like:
> `static lChar32 getReplacementChar(lUInt32 code, bool * ignore_if_not_found=NULL)`

Something like this: https://github.com/buggins/coolreader/pull/200/commits/49b0ada02dec0ebb646a4dcf88c3407ce5daa266 ?

virxkane:
> When you ignore it in the drawing with this PR as it is, do you still some blank (taken by the "?" which is not drawn) ?

Does this mean if a blank character is drawn?

poire-z:
> Something like this

Yep.

> Does this mean if a blank character is drawn?

Reformulating :) I wonder if that way, just not drawing it but still keeping it when measuring and positionning words (so, accounting for the width of the def_char='?', you would not get some unwanted blank space before or after the footnote number `, [2]`.
Measured: `,?[2]`
Drawn: `, [2]`
Expected: `,[2]`

virxkane:
> Reformulating :) I wonder if that way, just not drawing it but still keeping it when measuring and positionning words (so, accounting for the width of the def_char='?', you would not get some unwanted blank space before or after the footnote number

When we calculate the width of the character we call the getGlyphInfo() function, which in turn calls getCharIndex(), which we already changed so that the width of the 'WORD JOINER' character will be 0.

virxkane:
@poire-z I don't see any blank spaces:
Screenshot_20201122_190306-m

poire-z:
Ok, good. Then this PR looks fine.

virxkane:
@poire-z Thank you.

virxkane:
> PS: what does `Tym6epm` (pardon my cyrillic) means ? :)

In this book fragment? Character `s name.

virxkane:
> Unless it's ok to just flag it in here:
> https://github.com/buggins/coolreader/blob/b01ce22844524e379b94d2ef176353ed0bf94d17/crengine/src/lvtextfm.cpp#L1366-L1380
> 
> 
> This was so long ago - I'm not sure which is the best :)

With next patch (and without all commits in this PR) U+2060 also is not rendered:
```diff
diff --git a/crengine/src/lvtextfm.cpp b/crengine/src/lvtextfm.cpp
index 089f3e29a..4a78e6d35 100644
--- a/crengine/src/lvtextfm.cpp
+++ b/crengine/src/lvtextfm.cpp
@@ -1376,6 +1376,7 @@ public:
                         if ( c <= 0x2069 ) {
                             if ( c <= 0x202E ) m_flags[pos] = LCHAR_IS_TO_IGNORE;      // 202A>202E
                             else if ( c >= 0x2066 ) m_flags[pos] = LCHAR_IS_TO_IGNORE; // 2066>2069
+                            else if ( c == 0x2060 ) m_flags[pos] = LCHAR_IS_TO_IGNORE; // 2060
                         }
                     }
                     else if ( c <= 0x009F ) {
```
But might need to update comments?

> This was so long ago - I'm not sure which is the best :)

Perhaps need to mark this symbol as ignored at an earlier stage? Those is this patch enough? But I don't know if the line with the footnote links will break with this patch.

poire-z:
> But I don't know if the line with the footnote links will break with this patch.

It's only flagged at this point, so the WORDJOINER char is still given to libunibreak, so it should prevent a break as expected.

> Perhaps need to mark this symbol as ignored at an earlier stage? Those is this patch enough?

Re-reading the code: LCHAR_IS_TO_IGNORE is used:
A) to split measuring sequences in measureText() when measuring the width of all chars in a paragraph
B) in AddLine() (so, after we decided where to split current line) when making individual words

I don't know how WORDJOINER might affect Harfbuzz:
- if we flag it, we'll get 2 harfbuzz measuring runs
- if we don't flag it, we'll get only 1, and Harfbuzz might decide to shape the whole combo in a specific more accurate way than with 2 runs.

I think it was OK to flag LCHAR_IS_TO_IGNORE the ones I did (bidi markers, rare control chars), but flagging WORDJOINER (that's supposed to join words) would cause runs and words to no more be joined but be splitted :) 
So, may be what you did is the less risky option: we keep giving the most pure stuff to Harfbuzz, and we only deal with the case for FreeType/HB-light when no fallback font has a glyph, which is probably rare.

virxkane:
> Re-reading the code: LCHAR_IS_TO_IGNORE is used:

It will take me another couple of weeks to study the file "lvtextfm.cpp", which is why I am asking you.

> So, may be what you did is the less risky option: we keep giving the most pure stuff to Harfbuzz, and we only deal with the case for FreeType/HB-light when no fallback font has a glyph, which is probably rare.

Well, then I leave it as it is. Thanks.

PR #201, opened 2020-11-22
ONYX devices support updated.

virxkane:
* Update onyxsdk: source removed, but added reworked "onyxsdk-device-1.1.3" as bundle instead. 
This update allows implement dual-color backlight control, and also adds support for ONYX devices based on rk3368 (Rockchip 3368), msm8953 (Snapdragon 625), Snapdragon 660 platforms, for example, Poke 2, Poke 3, Nova 2, Nova 3, Note 2, Note 3, Note Air, Max 3, Max Lumi. But these devices have not been tested.
* Do not make a full screen refresh on eink devices when rendering a progress bar.
* Added an action to open the system dialog for setting the backlight brightness on ONYX BOOX devices.
* Added ability to adjust warm backlight on ONYX BOOX devices.
* For eink devices, added variant "1" for option "Full screen update interval".
* Fixed a bug on ONYX BOOX devices based on rk3288 that prevented updating the image on the screen.


virxkane:
Does anyone know what the `byPass(int count)` function does?

PR #204, opened 2020-12-05
Fix TryExec and Exec paths in Qt variant's desktop file

AndyMender:
Changes the `TryExec` and `Exec` paths in the desktop file of the Qt variant to fully qualified `/usr/bin/cr3`.


cr3qt/src/desktop/cr3.desktop
```diff
@@ -61,8 +61,8 @@ GenericName[zh_HK]=電子書閱讀器
 GenericName[zh_TW]=電子書閱讀器
 Comment=A cross-platform XML/CSS based eBook reader
 Comment[ru]=Кроссплатформенное чтение электронных книг, основанное на XML/CSS
-TryExec=cr3
-Exec=cr3 %F
+TryExec=/usr/bin/cr3
+Exec=/usr/bin/cr3 %F
```
virxkane, line 65: If I call `cmake` with argument `CMAKE_INSTALL_PREFIX=/usr/local` cr3.desktop will be invalid.

AndyMender: Fair point. Probably doesn't make sense to hard-code the entire path then. I'm closing the PR.

virxkane: But you can create the file `cr3.desktop.cmake` as a template and using [configure_file](https://cmake.org/cmake/help/latest/command/configure_file.html) will generate the file`cr3.desktop` where `CMAKE_INSTALL_PREFIX` will be replaced with the actual installation path.

AndyMender: I actually tested the original approach (just `cr3`) recently and it works well enough. I'm not sure why the patch was added to Fedora in the first place, but I removed it.

PR #212, opened 2021-01-05
HTML lists: support the 'reversed' attribute

ourairquality:
Support reversed list item counting.

Fix the count to advance irrespective of the list style.

Here's an example to illustrate reversed counting and some of the other fixes.

```
<ol reversed="reversed" start="3" style="list-style-type: upper-alpha;">
  <li> expect C</li>
  <li> expect B</li>
  <li> expect A</li>
  <li> expect 0</li>
  <li> expect -1</li>
  <li value="10"> expect J</li>
  <li style="list-style-type: lower-alpha"> expect i</li>
  <li style="list-style-type: none;"> expect </li>
  <li> expect G</li>
  <li style="list-style-type: lower-alpha"> expect f</li>
  <li style="list-style-type: square;"> expect square</li>
  <li> expect D</li>
  <li style="list-style-type: lower-alpha"> expect c</li>
</ol>
```


poire-z:
Looks simpler than github diff would show (used `git diff -b -U20` after applying).

But is this the correct behaviour, if I just add "reversed=true" to my OL, but no start= or value= ?:
<kbd>image</kbd>

https://www.w3.org/TR/html52/grouping-content.html#the-ol-element says:
> If the start attribute is present, user agents must parse it as an integer, in order to determine the attribute’s value. The default value, used if the attribute is missing or if the value cannot be converted to a number according to the referenced algorithm, is 1 if the element has no reversed attribute, and is **the number of child li elements** otherwise.

So, I guess it should show 7 6 5 4 3 2 1 on my screenshot?

ourairquality:
Yes, good point, the default start value needs to be set differently. Can look into that too.

poire-z:
Btw, you (or people doing the next release) will need to increment this because of the added attribute that will change IDs of the next attributes, which will cause those saved in existing book DOM caches to be invalid:
https://github.com/buggins/coolreader/blob/face2671e1690f56cbaaf97d390c7b9c4dcd1027/crengine/src/lvtinydom.cpp#L87-L88

poire-z (approved):
Looks allright and logical, and works fine on my few list item test cases.

crengine/src/lvtinydom.cpp
```diff
@@ -17553,21 +17553,69 @@ bool ldomNode::getNodeListMarker( int & counterValue, lString32 & marker, int &
...
+                }
+            }
+
             // The UL > LI parent-child chain may have had some of our Boxing elements inserted
             ldomNode * parent = getUnboxedParent();
-            counterValue = 0;
-            // See if parent has a 'start' attribute that overrides this 0
+
+            // See if parent has a 'reversed' attribute.
+            int increment = parent->getAttributeValue(attr_reversed).empty() ? +1 : -1;
```
poire-z, line 17571: A bit late, but:
https://html.spec.whatwg.org/multipage/grouping-content.html#attr-ol-reversed
> The reversed attribute is a boolean attribute. If present, it indicates that the list is a descending list (..., 3, 2, 1). If the attribute is omitted, the list is an ascending list (1, 2, 3, ...).
A number of attributes are boolean attributes. The presence of a boolean attribute on an element represents the true value, and the absence of the attribute represents the false value.

So, it seems `<ol reversed>` is valid - and your `parent->getAttributeValue(attr_reversed).empty() ` will match and you won't get things reversed. You could use: `parent->hasAttribute(attr_reversed) `
I don't know if `<ol reversed=false>` means it's not reversed and if more checks are needed.

(With all the _past HTML specs and what it meant at the time_ uncertainties :)

ourairquality: Thanks fixed in https://github.com/buggins/coolreader/pull/227

Tested a few web browsers and they appear to reverse the list irrespective of the value of the attribute, even if 'false'.

PR #215, opened 2021-01-08
HTML open-quote close-quote: correct internally encoded string length.

ourairquality:
The strings for open-quote and close-quote were not being copied fully
due to an off-by-one error in the encoded string length. This was
showing up as the closing quote being omitted.


poire-z:
I can confirm there is a regression with quotes (and possibly other stuff) between KOReader 2020.09 and 2020.10.
This PR _feels_ like it's not the right fix (but if it is, the comment is wrong, as a quote length is normally 2 chars).
Will investigate.

poire-z:
OK, looks like it is the right fix.
But the comment should say:
`// length+1 as expected with 's' by get_applied_content_property()`

ourairquality:
Updated the comments, thanks.

poire-z (approved):

PR #216, opened 2021-01-08
HTML docs: support the 'q' element, via the css files

ourairquality:
This can work now, so it could be supported when reading html docs.

### DIFF
--- a/android/res/raw/htm.css
+++ b/android/res/raw/htm.css
@@ -80,6 +80,9 @@ blockquote { display: block;
    margin-top: 0.5em; 
    margin-bottom: 0.5em; 
    } 
+/* No support for the "quotes:" property, these will use default quote chars */
+q::before               { content: open-quote; }
+q::after                { content: close-quote; }
 
 img { 
    margin: 0.5em; 
--- a/cr3gui/data/htm.css
+++ b/cr3gui/data/htm.css
@@ -244,3 +244,7 @@ table { font-size: 100% }
 td, th { text-indent: 0px; padding: 3px } 
 th { font-weight: bold; text-align: center } 
 table caption { text-indent: 0px; padding: 4px }
+
+/* No support for the "quotes:" property, these will use default quote chars */
+q::before               { content: open-quote; }
+q::after                { content: close-quote; }
--- a/cr3qt/data/htm.css
+++ b/cr3qt/data/htm.css
@@ -80,6 +80,9 @@ blockquote { display: block;
    margin-top: 0.5em; 
    margin-bottom: 0.5em; 
    } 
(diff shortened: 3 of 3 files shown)


virxkane:
@ourairquality 
Test android build, I think some improvements are needed:
Screenshot_1610464647-sz
Test file: [test1.zip](https://github.com/buggins/coolreader/files/5803014/test1.zip)

poire-z:
This issue (closing quotes not shown) should be solved by #215 (that I'll pick for KOReader).
(I won't pick this #216 as we don't update anymore this htm.css that we consider "old/obsolete".)

virxkane:
Well, he should have written about it.

virxkane:
Ok, it works.

ourairquality:
Also the quote is unexpected for english due to the default choice of the Russian hypernations, but I have not had time to look into that yet.

This PR seems ok, but yes more work is need, e.g. the text selection does not copy the pseudo elements, they do not work well with Read-Aloud etc. I am still trying to understand the code in the area and would like to get this working better to improve Read-Aloud for accessibility.

poire-z:
> the text selection does not copy the pseudo elements, they do not work well with Read-Aloud 

Just mentionning it: CSS `content:`'s content (and so open-quote/close-quote) are added when drawing text - but web browsers do not make part of the selected text that is returned for bookmarks, etc... (we keep them part of the rectangles when selecting/highlighting text around them this).
So, they are just cosmectics, not content.
(If you _really_ need them for ReadAloud, this most probably needs to be an option if you want to include them when fetching the text.)

ourairquality:
> > the text selection does not copy the pseudo elements, they do not work well with Read-Aloud
> 
> Just mentionning it: CSS `content:`'s content (and so open-quote/close-quote) are added when drawing text - but web browsers do not make part of the selected text that is returned for bookmarks, etc..

Yes I see that in Chrome and Firefox text selection. Fwiw they do copy the img alt text in text selection, and perhaps that is also 'added when drawing text'? Happy to add an option for it people don't want the pseudo elements in selections as a default, or perhaps it needs much more development and other HTML extension would be better used - anyway that is what I am exploring and trying to do.

Perhaps the following would be of interest, is notes web browser and screen reader support for CSS generated content, and there are a good number of combinations with support, and that seems a good precedent.
https://tink.uk/accessibility-support-for-css-generated-content/

I even see mention that Voice Over can give the size of a list at the start, but even being able to use ::before and ::after might be a big leap, and being able to read out the markers.

It also opens the pseudo elements up for search in CR, parts of the document that might have only been searchable visually can become searchable. Fwiw adding the img alt text as a child in CR makes it searchable, but translating that back to a point range is not yet supported.

PR #217, opened 2021-01-08
Distinct list circle and disc characters

ourairquality:
The source included commented out distinct list circle and disc characters which this PR enables. It had been using the same characters for both, and a different character. Could this be revisited now so that there are distinct circle and disc list labels available? Was there are lack of support in the fonts, or was it a presentation decision?

### DIFF
--- a/crengine/src/private/lvfreetypeface.cpp
+++ b/crengine/src/private/lvfreetypeface.cpp
@@ -836,10 +836,10 @@ bool LVFreeTypeFace::hbCalcCharWidth(LVCharPosInfo *posInfo, const LVCharTriplet
         }
 
     }
-    // Otherwise, use plain Freetype getGlyphInfo() which will check
+    // Otherwise, use plain Freetype getGlyphInfoSearch() which will check
     // again with this font, or the fallback one
     glyph_info_t glyph;
-    if ( getGlyphInfo(triplet.Char, &glyph, def_char, fallbackPassMask) ) {
+    if ( getGlyphInfoSearch(triplet.Char, &glyph, def_char, fallbackPassMask) ) {
         posInfo->offset = 0;
         posInfo->width = glyph.width;
         return true;
@@ -849,7 +849,7 @@ bool LVFreeTypeFace::hbCalcCharWidth(LVCharPosInfo *posInfo, const LVCharTriplet
 
 #endif  // USE_HARFBUZZ==1
 
-FT_UInt LVFreeTypeFace::getCharIndex(lUInt32 code, lChar32 def_char) {
+FT_UInt LVFreeTypeFace::getCharIndex(lUInt32 code, bool replace_missing, lChar32 def_char) {
     if (code == '\t')
         code = ' ';
     FT_UInt ch_glyph_index = FT_Get_Char_Index(_face, code);
@@ -863,7 +863,7 @@ FT_UInt LVFreeTypeFace::getCharIndex(lUInt32 code, lChar32 def_char) {
             FT_Select_Charmap(_face, FT_ENCODING_UNICODE);
         }
     }
-    if ( ch_glyph_index==0 ) {
+    if ( ch_glyph_index==0 && replace_missing ) {
         bool can_be_ignored = false;
         lUInt32 replacement = getReplacementChar( code, &can_be_ignored );
         if ( replacement )
@@ -1020,22 +1020,20 @@ void LVFreeTypeFace::setupHBFeatures()
 }
 #endif
 
-bool LVFreeTypeFace::getGlyphInfo(lUInt32 code, LVFont::glyph_info_t *glyph, lChar32 def_char, lUInt32 fallbackPassMask) {
+bool LVFreeTypeFace::getGlyphInfo(lUInt32 code, LVFont::glyph_info_t *glyph, lChar32 def_char,
+                                  lUInt32 fallbackPassMask, bool replace_missing) {
     //FONT_GUARD
-    int glyph_index = getCharIndex(code, 0);
+    int glyph_index = getCharIndex(code, replace_missing, def_char);
     if (glyph_index == 0) {
         LVFont *fallback = getFallbackFont(fallbackPassMask);
-        if (NULL == fallback) {
-            // No fallback
-            glyph_index = getCharIndex(code, def_char);
-            if (glyph_index == 0)
-                return false;
-        } else {
-            // Fallback
-            lUInt32 passMask = fallbackPassMask | _fallback_mask;
-            return fallback->getGlyphInfo(code, glyph, def_char, passMask);
-        }
+        if (!fallback)
+            return false;
+        return fallback->getGlyphInfo(code, glyph, def_char,
+                                      fallbackPassMask | _fallback_mask,
+                                      replace_missing);
     }
+
+    //FONT_GUARD
     int flags = FT_LOAD_DEFAULT;
     flags |= (!_drawMonochrome ? FT_LOAD_TARGET_NORMAL : FT_LOAD_TARGET_MONO);
     if (_hintingMode == HINTING_MODE_BYTECODE_INTERPRETOR) {
@@ -1133,6 +1131,14 @@ bool LVFreeTypeFace::getGlyphInfo(lUInt32 code, LVFont::glyph_info_t *glyph, lCh
     // a ceil()'ed value when considering negative numbers as some overflow,
     // which is good when we're using it for adding some padding.
     return true;
+}
+
+bool LVFont::getGlyphInfoSearch(lUInt32 code, LVFont::glyph_info_t *glyph, lChar32 def_char,
+                                lUInt32 fallbackPassMask) {
+    // Firstly try the font and the fallbacks list without character replacement,
+    // and then if necessary retry with character replacement.
+    return getGlyphInfo(code, glyph, def_char, fallbackPassMask, false) ||
+           getGlyphInfo(code, glyph, def_char, fallbackPassMask, true);
 }
 
 bool LVFreeTypeFace::checkFontLangCompat(const lString8 &langCode) {
@@ -1652,7 +1658,7 @@ lUInt16 LVFreeTypeFace::measureText(const lChar32 *text,
 #if (ALLOW_KERNING==1)
         if ( use_kerning && previous>0  ) {
             if ( ch_glyph_index==(FT_UInt)-1 )
-                ch_glyph_index = getCharIndex( ch, def_char );
+                ch_glyph_index = getCharIndex( ch, true, def_char );
             if ( ch_glyph_index != 0 ) {
                 FT_Vector delta;
                 error = FT_Get_Kerning( _face,          /* handle to face object */
@@ -1671,7 +1677,7 @@ lUInt16 LVFreeTypeFace::measureText(const lChar32 *text,
         int w = _wcache.get(ch);
         if ( w == CACHED_UNSIGNED_METRIC_NOT_SET ) {
             glyph_info_t glyph;
-            if ( getGlyphInfo( ch, &glyph, def_char, fallbackPassMask ) ) {
+            if ( getGlyphInfoSearch( ch, &glyph, def_char, fallbackPassMask ) ) {
                 w = glyph.width;
                 _wcache.put(ch, w);
             } else {
@@ -1682,7 +1688,7 @@ lUInt16 LVFreeTypeFace::measureText(const lChar32 *text,
         }
         if ( use_kerning ) {
             if ( ch_glyph_index==(FT_UInt)-1 )
-                ch_glyph_index = getCharIndex( ch, 0 );
+                ch_glyph_index = getCharIndex( ch, false);
             previous = ch_glyph_index;
         }
         widths[i] = prev_width + w + FONT_METRIC_TO_PX(kerning);
@@ -1764,21 +1770,19 @@ void LVFreeTypeFace::updateTransform() {
     //        }
 }
 
-LVFontGlyphCacheItem *LVFreeTypeFace::getGlyph(lUInt32 ch, lChar32 def_char, lUInt32 fallbackPassMask) {
+LVFontGlyphCacheItem *LVFreeTypeFace::getGlyph(lUInt32 ch, lChar32 def_char,
+                                               lUInt32 fallbackPassMask,
+                                               bool replace_missing) {
     //FONT_GUARD
-    FT_UInt ch_glyph_index = getCharIndex(ch, 0);
+    FT_UInt ch_glyph_index = getCharIndex(ch, replace_missing, def_char);
     if (ch_glyph_index == 0) {
         LVFont *fallback = getFallbackFont(fallbackPassMask);
-        if (NULL == fallback) {
-            // No fallback
-            ch_glyph_index = getCharIndex(ch, def_char);
-            if (ch_glyph_index == 0)
-                return NULL;
-        } else {
-            // Fallback
-            return fallback->getGlyph(ch, def_char, fallbackPassMask | _fallback_mask );
-        }
+        if (!fallback)
+            return NULL;
+        return fallback->getGlyph(ch, def_char, fallbackPassMask | _fallback_mask,
+                                  replace_missing);
     }
+
     LVFontGlyphCacheItem *item = _glyph_cache.get(ch);
     if (!item) {
         int rend_flags = FT_LOAD_RENDER | (!_drawMonochrome ? FT_LOAD_TARGET_LIGHT
@@ -1825,6 +1829,17 @@ LVFontGlyphCacheItem *LVFreeTypeFace::getGlyph(lUInt32 ch, lChar32 def_char, lUI
             _glyph_cache.put(item);
     }
     return item;
+}
+
+LVFontGlyphCacheItem *LVFont::getGlyphSearch(lUInt32 ch, lChar32 def_char,
+                                                     lUInt32 fallbackPassMask) {
+
+    // Firstly try the font and the fallbacks list without character replacement,
+    // and then if necessary retry with character replacement.
+    LVFontGlyphCacheItem *item = getGlyph(ch, def_char, fallbackPassMask, false);
+    if (item)
+        return item;
+    return getGlyph(ch, def_char, fallbackPassMask, true);
 }
 
 #if USE_HARFBUZZ == 1
@@ -1894,7 +1909,7 @@ int LVFreeTypeFace::getCharWidth(lChar32 ch, lChar32 def_char) {
     lUInt16 w = _wcache.get(ch);
     if (w == CACHED_UNSIGNED_METRIC_NOT_SET) {
         glyph_info_t glyph;
-        if (getGlyphInfo(ch, &glyph, def_char)) {
+        if (getGlyphInfoSearch(ch, &glyph, def_char)) {
             w = glyph.width;
         } else {
             w = 0;
@@ -1911,7 +1926,7 @@ int LVFreeTypeFace::getLeftSideBearing( lChar32 ch, bool negative_only, bool ita
     lInt16 b = _lsbcache.get(ch);
     if ( b == CACHED_SIGNED_METRIC_NOT_SET ) {
         glyph_info_t glyph;
-        if ( getGlyphInfo( ch, &glyph, '?' ) ) {
+        if ( getGlyphInfoSearch( ch, &glyph, '?' ) ) {
             b = glyph.originX;
         }
         else {
@@ -1931,7 +1946,7 @@ int LVFreeTypeFace::getRightSideBearing( lChar32 ch, bool negative_only, bool it
     lInt16 b = _rsbcache.get(ch);
     if ( b == CACHED_SIGNED_METRIC_NOT_SET ) {
         glyph_info_t glyph;
-        if ( getGlyphInfo( ch, &glyph, '?' ) ) {
+        if ( getGlyphInfoSearch( ch, &glyph, '?' ) ) {
             b = glyph.rsb;
         }
         else {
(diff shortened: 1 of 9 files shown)


poire-z:
I don't have the answer, and neither https://github.com/buggins/coolreader/commit/ec5ea6070

poire-z:
Although I can understand the choice :) these chars might be too fat in some/most fonts.

With KOReader's defaut fonts, currently | with this PR changes:
image - image

ourairquality:
Revised to use smaller characters, listing some options in the comments.

ourairquality:
Sorry some dithering. Kept the same 'disc' character which will have the same font coverage as before.

But the PR now proposes a smaller 'square' character, and a distinct 'circle' character. Updated the replacements so there is a fallback at least.

poire-z:
These new chars, when available in a same font, looks all quite nice and coherent:
FreeSans | FreeSerif
image - image

It happens that the 2 possible chars for square are present in most fonts - while the ones for circle and disc are not, and KOReader find them only in its lasts fallback fonts: FreeSans/FreeSerif.
U+2022 was found in most fonts.
Dunno what fallback fonts CoolReader ships, and if these will always be found in some ways.

But the fact that squares comes from the main font, and disc/circle from another, could cause these fluctuations in the list item indentation/aligmnents (but it's not worse as before, on the right :)
With this PR | before:
image image


I guess this may not be an issue, as most often, there are nested and will get some additional indentation from padding/margin which will make the difference not noticable.

ourairquality:
Made an error trying to include the 'circle' char in the fallback replacement, fixed.

Checked an Android coolreader build and the circle does not appear to be supported in any of the fonts, and there seems to be very limited support for the square symbol. Exploring further, want to look into the actual fonts files and see what glyphs they do support, see if something is broken in the logic etc. Can we bundle in some extra font etc.

ourairquality:
Changed the 'circle' character to ```25E6``` as this is included in FreeSans and seems to have better coverage in general.

The fallback font is not being used for characters in the replacement set because the logic firstly checks for a replacement character when a character is not found in the primary font and before searching the fallback fonts. If these codes are remove from the replacement set - commented out - then they can be supplied from a fallback font.

Would this have been the intention, to give preference to a replacement character over a matching code point in a fallback font?

Or would people like this logic to be flipped to only search for a replacement character if there is no match in both the primary font or any of the fallback fonts?

virxkane:
> The fallback font is not being used for characters in the replacement set because the logic firstly checks for a replacement character when a character is not found in the primary font and before searching the fallback fonts. If these codes are remove from the replacement set - commented out - then they can be supplied from a fallback font.

In fact, it depends on the text shaping mode (or kerning mode in KOReader), for SHAPING_MODE_HARFBUZZ this is not the case: if there is no such glyph in the main font, but a fallback font is specified, then its rendering is delegated to the fallback font, if not set, then replacement characters are used only in this case.

> Or would people like this logic to be flipped to only search for a replacement character if there is no match in both the primary font or any of the fallback fonts?

Who knows? My opinion is yes.

poire-z:
(CR and KR font code is a bit different, at least in different files, but may be the logic is the same on this point, haven't checked).
If I understand our current code right:
HARFBUZZ FULL: getReplacementChar/filterChar is only called on the last fallback font, so we get the replacement char from the last fallback font (if there is no replacement, we get the def_char from the last fallback font)
OTHERS: getReplacementChar is called on the first font, so if there's a replacement, we use the one from the first font (if there is no replacement, we go look in fallback fonts, and ultimately use the def_char from the last fallback font - as we keep provided def_char=0 until the last fallback font).

I think having OTHERS behave as HARFBUZZ FULL is just  matter of:
```diff
     FT_UInt getCharIndex( lChar32 code, lChar32 def_char ) {
         if ( code=='\t' )
             code = ' ';
         FT_UInt ch_glyph_index = FT_Get_Char_Index( _face, code );
         if ( ch_glyph_index==0 && code >= 0xF000 && code <= 0xF0FF) {
             // If no glyph found and code is among the private unicode
             // area classically used by symbol fonts (range U+F020-U+F0FF),
             // try to switch to FT_ENCODING_MS_SYMBOL
             if (!FT_Select_Charmap(_face, FT_ENCODING_MS_SYMBOL)) {
                 ch_glyph_index = FT_Get_Char_Index( _face, code );
                 // restore unicode charmap if there is one
                 FT_Select_Charmap(_face, FT_ENCODING_UNICODE);
             }
         }
-        if ( ch_glyph_index==0 ) {
+        if ( ch_glyph_index==0 && def_char ) { // only look for replacement when def_char != 0
             bool can_be_ignored = false;
             lUInt32 replacement = getReplacementChar( code, &can_be_ignored );
             if ( replacement )
                 ch_glyph_index = FT_Get_Char_Index( _face, replacement );
             if ( ch_glyph_index==0 && def_char && !can_be_ignored ) {
                 // if neither the index of this character nor the index of the replacement character is found,
                 // and if this character can be safely ignored,
                 // we simply skip it so as not to draw the unnecessary replacement character.
                 ch_glyph_index = FT_Get_Char_Index( _face, def_char );
             }
         }
         return ch_glyph_index;
     }
```

(Best would have been, if glyph not found in any font, to return back to the first font and use getReplacementChar() from it - but that would need more rework on the already fragile font code :/ so best to not go at best :)

virxkane:
@poire-z Perhaps this is a solution, but ugly, looking at the arguments of the function, it is completely unclear that if def_char is 0, then a replacement search for a character will not be performed if there is no glyph in this font.
May be something like this:
```diff
diff --git a/crengine/src/private/lvfreetypeface.cpp b/crengine/src/private/lvfreetypeface.cpp
index 02758ddb9..25d8bc0dd 100644
--- a/crengine/src/private/lvfreetypeface.cpp
+++ b/crengine/src/private/lvfreetypeface.cpp
@@ -845,7 +845,7 @@ bool LVFreeTypeFace::hbCalcCharWidth(LVCharPosInfo *posInfo, const LVCharTriplet
 
 #endif  // USE_HARFBUZZ==1
 
-FT_UInt LVFreeTypeFace::getCharIndex(lUInt32 code, lChar32 def_char) {
+FT_UInt LVFreeTypeFace::getCharIndex(lUInt32 code, bool replace_missing, lChar32 def_char) {
     if (code == '\t')
         code = ' ';
     FT_UInt ch_glyph_index = FT_Get_Char_Index(_face, code);
@@ -859,7 +859,7 @@ FT_UInt LVFreeTypeFace::getCharIndex(lUInt32 code, lChar32 def_char) {
             FT_Select_Charmap(_face, FT_ENCODING_UNICODE);
         }
     }
-    if ( ch_glyph_index==0 ) {
+    if ( ch_glyph_index==0 && replace_missing ) {
         bool can_be_ignored = false;
         lUInt32 replacement = getReplacementChar( code, &can_be_ignored );
         if ( replacement )
@@ -1018,12 +1018,12 @@ void LVFreeTypeFace::setupHBFeatures()
 
 bool LVFreeTypeFace::getGlyphInfo(lUInt32 code, LVFont::glyph_info_t *glyph, lChar32 def_char, lUInt32 fallbackPassMask) {
     //FONT_GUARD
-    int glyph_index = getCharIndex(code, 0);
+    int glyph_index = getCharIndex(code, false);
     if (glyph_index == 0) {
         LVFont *fallback = getFallbackFont(fallbackPassMask);
         if (NULL == fallback) {
             // No fallback
-            glyph_index = getCharIndex(code, def_char);
+            glyph_index = getCharIndex(code, true, def_char);
             if (glyph_index == 0)
                 return false;
         } else {
@@ -1648,7 +1648,7 @@ lUInt16 LVFreeTypeFace::measureText(const lChar32 *text,
 #if (ALLOW_KERNING==1)
         if ( use_kerning && previous>0  ) {
             if ( ch_glyph_index==(FT_UInt)-1 )
-                ch_glyph_index = getCharIndex( ch, def_char );
+                ch_glyph_index = getCharIndex( ch, true, def_char );
             if ( ch_glyph_index != 0 ) {
                 FT_Vector delta;
                 error = FT_Get_Kerning( _face,          /* handle to face object */
@@ -1678,7 +1678,7 @@ lUInt16 LVFreeTypeFace::measureText(const lChar32 *text,
         }
         if ( use_kerning ) {
             if ( ch_glyph_index==(FT_UInt)-1 )
-                ch_glyph_index = getCharIndex( ch, 0 );
+                ch_glyph_index = getCharIndex( ch, false);
             previous = ch_glyph_index;
         }
         widths[i] = prev_width + w + FONT_METRIC_TO_PX(kerning);
@@ -1762,12 +1762,12 @@ void LVFreeTypeFace::updateTransform() {
 
 LVFontGlyphCacheItem *LVFreeTypeFace::getGlyph(lUInt32 ch, lChar32 def_char, lUInt32 fallbackPassMask) {
     //FONT_GUARD
-    FT_UInt ch_glyph_index = getCharIndex(ch, 0);
+    FT_UInt ch_glyph_index = getCharIndex(ch, false);
     if (ch_glyph_index == 0) {
         LVFont *fallback = getFallbackFont(fallbackPassMask);
         if (NULL == fallback) {
             // No fallback
-            ch_glyph_index = getCharIndex(ch, def_char);
+            ch_glyph_index = getCharIndex(ch, true, def_char);
             if (ch_glyph_index == 0)
                 return NULL;
         } else {
@@ -2315,7 +2315,7 @@ int LVFreeTypeFace::DrawTextString(LVDrawBuf *buf, int x, int y, const lChar32 *
             ch = UNICODE_SOFT_HYPHEN_CODE;
             isHyphen = 0;
         }
-        FT_UInt ch_glyph_index = getCharIndex( ch, def_char );
+        FT_UInt ch_glyph_index = getCharIndex( ch, true, def_char );
         int kerning = 0;
 #if (ALLOW_KERNING==1)
         if ( use_kerning && previous>0 && ch_glyph_index>0 ) {
diff --git a/crengine/src/private/lvfreetypeface.h b/crengine/src/private/lvfreetypeface.h
index ee80a1503..f53f4d9b9 100644
--- a/crengine/src/private/lvfreetypeface.h
+++ b/crengine/src/private/lvfreetypeface.h
@@ -382,7 +382,7 @@ public:
     virtual void Clear();
 protected:
     void updateTransform();
-    FT_UInt getCharIndex(lUInt32 code, lChar32 def_char);
+    FT_UInt getCharIndex(lUInt32 code, bool replace_missing, lChar32 def_char = 0);
 #if USE_HARFBUZZ==1
     LVFontGlyphCacheItem *getGlyphByIndex(lUInt32 index);
     lChar32 filterChar(lChar32 code, lChar32 def_char=0); 
```

virxkane:
@ourairquality Can you test the patch from @poire-z or mine?

virxkane:
@ourairquality I'll try again. Can you apply patch above to this PR?

poire-z:
About the latest character choices for disc/circle/square:
- they are all nice and coherent in sizes when found in a font
- disc is found in mostly all fonts, circle & square are rarely found
- so, circle & square will most often be picked from a fallback font, and will be coherent - but disc will sometimes not be (but rarely the difference is striking).
- when picked from different fonts, the difference in char width can be noticable (so, the indentation can change, as seen below).

So, rather quite OK.
Here's how it looks with various main fonts (with KOReader's default fallback fonts):
image

(When no fallback font, disck being found as a small bullet, and the letter `o` used as fallback for the circle, have quite a large difference in size, and it's ugly :) 

Something to may be quickly look at, if we trust our fallback fonts, is to use a less commonly found one than U+2022, so it gets as much chances to not be found as the 2 others, so that all 3 comes from the fallback font?
Although we mostly always see U+2022 in books - so it might just be better to keep using it so it comes from the main font - and assume the UL/LI padding/indentation will make the difference in char width not noticable.

ourairquality:
> @ourairquality I'll try again. Can you apply patch above to this PR and test?

Yes the patch works, on a simple test, with a single fallback font face.

Could anyone confirm my attempt to understand this code?

* The code appears to use a mask set of flags to limit the set of fallback font faces to consider when retrying, the fallbackPassMask. 

* LVFreeTypeFace::getFallbackFont() builds and/or walks a linked list of font fallback fonts, each with a different mask bit set, and returns the first that is not in the fallbackPassMask set.

* LVFreeTypeFace::getGlyphInfo() now looks up the char index without replacement, and if that fails it recursively retries with the failing font added to the mask set, so it effectively walks the list of fallback fonts. If this walk completes without finding a glyph for the char index it then retries using character replacements, but at this depth of the recursion the initial font and all other fallbacks have been added to the mask set so only the last fallback font remains and the replacement character will be take from that.

So it seems to work, but it does not seem ideal as the replacement character is taken from the last fallback font and not from the initial font - or also likely I got lost trying to understand this code!

poire-z:
CR and KR diverge a bit on this, but I think you understood it right.

> So it seems to work, but it does not seem ideal as the replacement character is taken from the last fallback font and not from the initial font - or also likely I got lost trying to understand this code!

Yes, wrote about that above:
> (Best would have been, if glyph not found in any font, to return back to the first font and use getReplacementChar() from it - but that would need more rework on the already fragile font code :/ so best to not go at best :)

virxkane:
@ourairquality 
> Could anyone confirm my attempt to understand this code?
...

Yes, that's right, that's how it works in CoolReader.

> So it seems to work, but it does not seem ideal as the replacement character is taken from the last fallback font and not from the initial font - or also likely I got lost trying to understand this code!

Yes, that's right, it is.

Now, it would be nice if you added a commit with this patch to this pull request so that we don't have to.

ourairquality:
I am having a go at improving this to retry all the fonts with replacement.

WIP
```
bool LVFreeTypeFace::getGlyphInfo(lUInt32 code, LVFont::glyph_info_t *glyph, lChar32 def_char, lUInt32 fallbackPassMask, bool replace_missing, bool topLevel) {
    //FONT_GUARD
    int glyph_index;

    if (replace_missing)
        glyph_index = getCharIndex(code, true, def_char);
    else
        glyph_index = getCharIndex(code, false);

    if (glyph_index == 0) {
        LVFont *fallback = getFallbackFont(fallbackPassMask);
        if (fallback == NULL)
            return false;
        // Fallback
        lUInt32 passMask = fallbackPassMask | _fallback_mask;
        bool success = fallback->getGlyphInfo(code, glyph, def_char, passMask, replace_missing, false);
        if (success)
            return true;
        if (!topLevel || replace_missing)
            return false;
        // Retry with replacement
        return getGlyphInfo(code, glyph, def_char, fallbackPassMask, true, false);
    }
...
```

Also might it be possible to simply the change to ```getCharIndex()``` to enable replacements when the def_char is defined and thus eliminate the new ```replace_missing``` argument? Would there be cases where replacement would be needed without a def_char, or visa_versa? If a def_char is always supplied for the same reasons as allowing replacement then it could be simplified a little?

crengine/src/private/lvfreetypeface.cpp
```diff
@@ -1019,22 +1019,31 @@ void LVFreeTypeFace::setupHBFeatures()
...
 #endif
 
-bool LVFreeTypeFace::getGlyphInfo(lUInt32 code, LVFont::glyph_info_t *glyph, lChar32 def_char, lUInt32 fallbackPassMask) {
+bool LVFreeTypeFace::getGlyphInfo(lUInt32 code, LVFont::glyph_info_t *glyph, lChar32 def_char,
+                                  lUInt32 fallbackPassMask, bool replace_missing, bool topLevel) {
     //FONT_GUARD
-    int glyph_index = getCharIndex(code, 0);
+    int glyph_index;
+
+    if (replace_missing)
```
virxkane, line 1027: This 4 lines can be replaced with with one:
```cpp
glyph_index = getCharIndex(code, replace_missing, def_char);
```

crengine/src/private/lvfreetypeface.cpp
```diff
@@ -1019,22 +1019,31 @@ void LVFreeTypeFace::setupHBFeatures()
...
+        glyph_index = getCharIndex(code, true, def_char);
+    else
+        glyph_index = getCharIndex(code, false);
```
virxkane, line 1030: This 4 lines can be replaced with with one:
```cpp
glyph_index = getCharIndex(code, replace_missing, def_char);
```

virxkane:
> Also might it be possible to simply the change to `getCharIndex()` to enable replacements when the def_char is defined and thus eliminate the new `replace_missing` argument? Would there be cases where replacement would be needed without a def_char, or visa_versa? If a def_char is always supplied for the same reasons as allowing replacement then it could be simplified a little?

Yes it is, this argument is added to improve the readability of the code.

crengine/src/private/lvfreetypeface.cpp
```diff
@@ -1020,22 +1020,25 @@ void LVFreeTypeFace::setupHBFeatures()
...
-            glyph_index = getCharIndex(code, def_char);
-            if (glyph_index == 0)
-                return false;
-        } else {
-            // Fallback
-            lUInt32 passMask = fallbackPassMask | _fallback_mask;
-            return fallback->getGlyphInfo(code, glyph, def_char, passMask);
-        }
+        if (fallback == NULL)
+            return false;
```
virxkane, line 1030: On the first call in the recursion chain, since replace_missing is false by default, we don't get a replacement char at all.

ourairquality:
Thanks again for the good review, and I missed that case. PTAL

virxkane:
> Thanks again for the good review, and I missed that case.

I just could not help but note the obvious shortcomings, but in fact, the point is different.

> I am having a go at improving this to retry all the fonts with replacement.

What I see in the code now (https://github.com/buggins/coolreader/pull/217/commits/2ac0f1b944884d43473e6d07d9e524e2f8570eb9):
`LVFreeTypeFace::getGlyphInfo`:
Let's say that in the main font (the first function call in the recursion chain), there is no symbol we need, and there is a fallback font:
* Then we try to get the character from the fallback font without using the character replacement list.
* If there is no such character in the fallback font, we again try to get it from the same font (main), but we already allow to use the character replacement list (`replace_missing=true`), but if it is also not in the replacement list or there is no replacement character either, then the replacement character is substituted for it (`def_char`).
* The next fallback font will not be used, since after calling `ch_glyph_index = getCharIndex (ch, true, def_char);` `ch_glyph_index` will not be zero.

As a result, if a symbol is not found in the main font, we do not go through the entire list of fallback fonts and almost completely break the "multiple fallback fonts" feature.

Correct me if I'm wrong.

In addition, it is not very nice to use a direct indication of the position of the function in the call stack (the topLevel argument) in a recursive function.

ourairquality (commented):
> > Thanks again for the good review, and I missed that case.
> 
> * Then we try to get the character from the fallback font without using the character replacement list.

It calls itself recursively to try the next fallback fonts in the list until it gets to the end of that list, so checks all the fallback fonts without replacement. That was the intention anyway, the ```fallback && fallback->getGlyphInfo()``` path, when that call to ```getGlyphInfo()``` returns with false it has checked all the rest of the fallback fonts in that list too.

> * If there is no such character in the fallback font, we again try to get it from the same font (main), but we already allow to use the character replacement list (`replace_missing=true`), but if it is also not in the replacement list or there is no replacement character either, then the replacement character is substituted for it (`def_char`).

This is also searched recursively, the same path, search all the fallback fonts in the list. So when if returns to the top level it has search them all.

> * The next fallback font will not be used, since after calling `ch_glyph_index = getCharIndex (ch, true, def_char);` `ch_glyph_index` will not be zero.

They will all have been searched due to the recursive search. That was the intention anyway.
 
> In addition, it is not very nice to use a direct indication of the position of the function in the call stack (the topLevel argument) in a recursive function.

Yes, in other languages I would have used block local functions and closures etc. C++ might be capable of that now, but such usage did not seem in-style with the rest of the CR code. If I have missed a C++ pattern that is better to use here then could you please point me to that, or perhaps just clean if up to suit as a follow up.

I did have a go at converting this to a loop, but the fallback fonts are in a linked list and this search is started from a method on the first, so this PR seemed the smallest change.

virxkane:
> It calls itself recursively to try the next fallback fonts in the list until it gets to the end of that list, so checks all the fallback fonts without replacement. That was the intention anyway, the `fallback && fallback->getGlyphInfo()` path, when that call to `getGlyphInfo()` returns with false it has checked all the rest of the fallback fonts in that list too.

Yes, indeed it is.

When I have free time, I'll see how it can be simplified and optimized.

virxkane:
@ourairquality Checked the second commit. Works correctly, but I simplified a little by removing the "replace_missing" and "topLevel" arguments and adding a "pass" argument that represents the pass number of the glyph search. If the value is "0", the search is performed without replacing the missing characters, with "1" - with the replacement. This required a small change in the handling of fallback fonts masks.
I hope this doesn't complicate the already rather convoluted code for working with fonts.
```cpp
bool LVFreeTypeFace::getGlyphInfo(lUInt32 code, LVFont::glyph_info_t *glyph, lChar32 def_char, lUInt32 fallbackPassMask, lUInt32 pass) {
    //FONT_GUARD
    FT_UInt glyph_index = getCharIndex(code, pass > 0, def_char);
    if (glyph_index == 0) {
        // This can only happen if the glyph is not found and replacement of the missing character is not permitted, or if the missing character can be safely ignored.
        LVFont *fallback = getFallbackFont(fallbackPassMask);
        if (fallback) {
            bool res = fallback->getGlyphInfo(code, glyph, def_char, fallbackPassMask | _fallback_mask, pass);
            if (res)
                return res;
        }
        if ( fallbackPassMask != 0 || pass > 0 ) {
            // go back one level up in recursion
            return false;
        }
        // Second pass with with replacement
        return getGlyphInfo(code, glyph, def_char, fallbackPassMask, 1);
    }
   ...
}
```
[20210120-two-pass-glyph-lookup.zip](https://github.com/buggins/coolreader/files/5841357/20210120-two-pass-glyph-lookup.zip)
Please check this patch and if you find it works and if you don't mind maybe replace commit 2ac0f1b944884d43473e6d07d9e524e2f8570eb9 by this patch?

ourairquality:
Moving the 'replace-missing' flag into the fallback masks appears to make this code even more difficult to understand (to me anyway) - it would appear to make it even harder for someone to unwind that state and to rewrite it in a functional style.

The update includes an alternative, removing the 'top-level' flag and effectively moving that it into a new higher level function which calls the existing function without and then with character replacement. The replacement argument is now also constant in a recursive search.

Just land a fix that you are happy with - it does not seem worth wasting time overthinking this one.

buggins:
reverted

PR #218, opened 2021-01-10
Android: workaround fallback font face(s) conflict

ourairquality:
The Android reader uses both PROP_FALLBACK_FONT_FACE and
PROP_FALLBACK_FONT_FACES and sets PROP_FALLBACK_FONT_FACES empty if
PROP_FALLBACK_FONT_FACE is defined. This could result in a single face
being set and then the faces being set to an empty set clearing the
initial face set and the fallback font setting not working.

Removing PROP_FALLBACK_FONT_FACES in this case does not solve this
problem as it will be defaulted to a hard coded list and override the
UI fallback font face setting.

The workaround here is to copy the PROP_FALLBACK_FONT_FACE into the
PROP_FALLBACK_FONT_FACES when set.

A cleaner fix would appear to be to remove PROP_FALLBACK_FONT_FACE and just have a single font placed in PROP_FALLBACK_FONT_FACES where that was the intention? The front end would then be responsible to managing a list of fallback fonts vs a single fallback font.

### DIFF
--- a/android/src/org/coolreader/crengine/ReaderView.java
+++ b/android/src/org/coolreader/crengine/ReaderView.java
@@ -2629,6 +2629,14 @@ public class ReaderView implements android.view.SurfaceHolder.Callback, Settings
 		props.remove(PROP_EMBEDDED_FONTS);
 		props.remove(PROP_REQUESTED_DOM_VERSION);
 		props.remove(PROP_RENDER_BLOCK_RENDERING_FLAGS);
+
+		// If a fallback font face is defined then set it in the font
+		// faces, otherwise the font faces may override the font face
+		// depending on which is applied first.
+		String fallbackFontFace = props.getProperty(PROP_FALLBACK_FONT_FACE);
+		if (fallbackFontFace.length() > 0)
+			props.setProperty(PROP_FALLBACK_FONT_FACES, fallbackFontFace);
+
 		BackgroundThread.ensureBackground();
 		log.v("applySettings()");
 		boolean isFullScreen = props.getBool(PROP_APP_FULLSCREEN, false);


virxkane:
> The Android reader uses both PROP_FALLBACK_FONT_FACE and
> PROP_FALLBACK_FONT_FACES and sets PROP_FALLBACK_FONT_FACES empty if
> PROP_FALLBACK_FONT_FACE is defined.

Can you point me to where this is happening?

> The workaround here is to copy the PROP_FALLBACK_FONT_FACE into the
> PROP_FALLBACK_FONT_FACES when set.

This is invalid workaround.

> A cleaner fix would appear to be to remove PROP_FALLBACK_FONT_FACE and just have a single font placed in PROP_FALLBACK_FONT_FACES where that was the intention?

Yes, planed to remove PROP_FALLBACK_FONT_FACE and add to UI possibility to manage PROP_FALLBACK_FONT_FACES, i.e. set **multiple** fallback fonts.

ourairquality:
> > The Android reader uses both PROP_FALLBACK_FONT_FACE and
> > PROP_FALLBACK_FONT_FACES and sets PROP_FALLBACK_FONT_FACES empty if
> > PROP_FALLBACK_FONT_FACE is defined.
> 
> Can you point me to where this is happening?

Perhaps that was not an accurate description, sorry. Rather it is what the debug trace showed, a single font being set, and then an empty set being set. This might have been caused by ```applyDefaultFallbackFontList```, not sure, but perhaps none of the default fonts were valid.

In any case there are two paths conflicting to set the fallback fonts.
 
> > The workaround here is to copy the PROP_FALLBACK_FONT_FACE into the
> > PROP_FALLBACK_FONT_FACES when set.
> 
> This is invalid workaround.

Ok. What is the expect behavior given the conflicting settings? Should the default fallback set still be applied, possibly overriding the UI selection, possibly overriding it with an empty set? 
 
> > A cleaner fix would appear to be to remove PROP_FALLBACK_FONT_FACE and just have a single font placed in PROP_FALLBACK_FONT_FACES where that was the intention?
> 
> Yes, planed to remove PROP_FALLBACK_FONT_FACE and add to UI possibility to manage PROP_FALLBACK_FONT_FACES, i.e. set **multiple** fallback fonts.

Could I go ahead and remove PROP_FALLBACK_FONT_FACE and keep the current UI settings for now, so the list is filled with just the one fallback font? This would be effectively the same behavior as this proposed PR, but perhaps moving in the desired direction rather than a hack workaround.

virxkane:
So, OK, but settting fallback font (PROP_FALLBACK_FONT_FACE) not replaced totally fallback fonts set (PROP_FALLBACK_FONT_FACES):
https://github.com/buggins/coolreader/blob/face2671e1690f56cbaaf97d390c7b9c4dcd1027/crengine/src/lvdocview.cpp#L6606-L6621
https://github.com/buggins/coolreader/blob/face2671e1690f56cbaaf97d390c7b9c4dcd1027/crengine/src/private/lvfreetypefontman.cpp#L41-L64
https://github.com/buggins/coolreader/blob/face2671e1690f56cbaaf97d390c7b9c4dcd1027/crengine/src/private/lvfreetypefontman.cpp#L66-L103

I think you end up with an empty set of fallback fonts, around here:
https://github.com/buggins/coolreader/blob/face2671e1690f56cbaaf97d390c7b9c4dcd1027/android/src/org/coolreader/crengine/BaseActivity.java#L1661-L1686

virxkane:
> Could I go ahead and remove PROP_FALLBACK_FONT_FACE and keep the current UI settings for now, so the list is filled with just the one fallback font?

No, this is a very bad idea: the multiple font fallback feature will be completely disabled!

We must implement in frontend (OptionDialog.java) the ability to select multiple fonts, not just one for PROP_FALLBACK_FONT_FACES. Something like an editable list: add a font (select from the list), remove a font.

ourairquality:
Closing for now, but shall explore further. Fwiw the cr3qt front end has a similar problem in that the UI fallback font is never used as the update is always followed by any update of the default fallback faces list.

PR #219, opened 2021-01-10
Android: default the hyphenation dictionary to 'algorithm'

ourairquality:
The language is currently derived from the hyphenation dictionary so having that defaulted to Russian defaulted the language to Russian, and that in turn affected the symbols chosen for quotes etc. It would have seemed better to default from the locale, but that would want to work for the 'algorithm' setting too, so this seems a start. There is an interesting option to use the books language, but I can't see that actually working, but perhaps something to look into. People might fairly consider this PR EN centric, but perhaps this is at least somewhere to discuss making better use of the locale and how that should default settings.

### DIFF
--- a/android/src/org/coolreader/crengine/BaseActivity.java
+++ b/android/src/org/coolreader/crengine/BaseActivity.java
@@ -1900,7 +1900,7 @@ public class BaseActivity extends Activity implements Settings {
 
 			props.setProperty(ReaderView.PROP_MIN_FILE_SIZE_TO_CACHE, "100000");
 			props.setProperty(ReaderView.PROP_FORCED_MIN_FILE_SIZE_TO_CACHE, "32768");
-			props.applyDefault(ReaderView.PROP_HYPHENATION_DICT, Engine.HyphDict.RUSSIAN.toString());
+			props.applyDefault(ReaderView.PROP_HYPHENATION_DICT, Engine.HyphDict.ALGORITHM.toString());
 			props.applyDefault(ReaderView.PROP_APP_FILE_BROWSER_SIMPLE_MODE, "0");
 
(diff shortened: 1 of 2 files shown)


virxkane:
> It would have seemed better to default from the locale, but that would want to work for the 'algorithm' setting too, so this seems a start.

Maybe so? Make by default the system locale.
Few users go into the settings to change the hyphenation dictionary, so if you set the default algorithmic hyphenation, the resulting hyphenation may be incorrect for some languages.

ourairquality:
> > It would have seemed better to default from the locale, but that would want to work for the 'algorithm' setting too, so this seems a start.
> 
> Maybe so? Make by default the system locale.
> Few users go into the settings to change the hyphenation dictionary, so if you set the default algorithmic hyphenation, the resulting hyphenation may be incorrect for some languages.

Ok, let me look into it further.

I would like to see it follow the document language too as a default, so use the document language if that is defined otherwise use the locale, then perhaps have a manual override in the UI for documents with incorrect language settings or to change the default?

For HTML the language can change within a scope and I presume the hyphenation should change with that, so when the document is wanting to dynamically change the language it seems somewhat backwards to be using a UI hyphenation dictionary setting to control the language.

virxkane:
> For HTML the language can change within a scope and I presume the hyphenation should change with that, so when the document is wanting to dynamically change the language it seems somewhat backwards to be using a UI hyphenation dictionary setting to control the language.

If it is intended to use several languages in one document, then, as far as I know, this has already been implemented.
https://github.com/buggins/coolreader/pull/142
https://github.com/koreader/crengine/pull/337

virxkane:
> There is an interesting option to use the books language, but I can't see that actually working, but perhaps something to look into

So, may be you fix this feature?
And then this option can be set by default.

ourairquality:
Closing this for now. Will follow with a new PR if a solution is developed.

PR #222, opened 2021-01-10
freetype: update to 2.10.4

ourairquality:
The freetype website claims a 'severe vulnerability' and suggests 'users should update immediately'. Not sure if this path affects coolreader, but why not update.


ourairquality:
> Could you remove the "thirdparty/freetype/docs/reference" directory in this update to reduce the size of the patch?

The size of the patch should not be an issue, it does not need to be reviewed file by file, the files are an exact copy from the distribution, including docs/reference which was already there. Perhaps you would like to just verify that the files are an exact copy from the distribution, and that would be enough, plus keep a look out for any reported regressions.

virxkane:
> The size of the patch should not be an issue, it does not need to be reviewed file by file, the files are an exact copy from the distribution, including docs/reference which was already there. Perhaps you would like to just verify that the files are an exact copy from the distribution, and that would be enough, plus keep a look out for any reported regressions.

Yes, of course. I am asking you to do this simply so that the resulting size of the "thirdparty" directory does not grow too much.

PR #225, opened 2021-01-12
HTML: emit img alt text in text selections and read-aloud

ourairquality:
This is adequate to get the text selection returning the img alt text, and that works with the Read-Aloud feature where it can be very useful.

Seems a bit of a hack, adding a child text element to an img element, but when you consider that the ::before and ::after pseudo elements can be used on the img element and are already added as child elements, and that these are to be emitted around the alt text, this approach is much simpler than having all the users include special cases to emit the alt text out of order with the pseudo elements (I tried that first and it was a big change that I never completed!)

This PR plays well with a WIP patch to get the pseudo elements emitted in text selection and thus for Read-Aloud.

If people have other suggestions then help would be appreciated?

### DIFF
--- a/crengine/src/lvtinydom.cpp
+++ b/crengine/src/lvtinydom.cpp
@@ -5236,6 +5236,18 @@ void ldomElementWriter::onBodyEnter()
                 }
             }
         }
+
+        // Add img alt text as a child. This will only be used by the
+        // graphical renderer if there is no image, but is needed for the text
+        // selection which is in turn used by Read-Aloud. This approach works
+        // simply with ::before and ::after pseudo elements which should be
+        // applied when the alternate text is emitted.
+        if ( _element->getNodeId() == el_img ) {
+            lString32 alt = _element->getAttributeValue( attr_alt );
+            if (!alt.empty())
+                _element->insertChildText(alt);
+        }
+
         _isBlock = isBlockNode(_element);
         // If initNodeStyle() has set "white-space: pre" or alike, update _flags
         if ( _element->getStyle()->white_space >= css_ws_pre_line) {


poire-z:
> Seems a bit of a hack

Indeed. Looks dubious to me. You don't know (and neither do I) how the rest of the code will behave with IMG having a text child.

Would feel better to have the text selection, when accumulating text content when walking nodes, when it meets a IMG, if it has a ALT or TITLE attribute, to just add their text to the accumulated text content. (And so with CSS content: if you really want them).
Rather than infecting the DOM :)

ourairquality:
> Would feel better to have the text selection, when accumulating text content when walking nodes, when it meets a IMG, if it has a ALT or TITLE attribute, to just add their text to the accumulated text content. (And so with CSS content: if you really want them).
> Rather than infecting the DOM :)

Yes that was my first attempt, however collecting the alt text in the correct order when walking the DOM, between potential ::before and ::after elements (that are already being added to the DOM), added far more complexity to the selection code paths.

ourairquality:
Chrome handles the ::before and ::after elements on the img alt text, and Firefox the ::after element. Many screen readers support CSS content.

It is easy getting the text collector above working, easy to also emit the list markers and ::before and ::after elements there too, but try getting all the text walking paths to work, e.g 'next sentence' 'prev sentence' etc - that is where it becomes clear that the text needs to be organised in order to simplify the text walking.

Screen readers augment the content, and might need to do so on content not specifically design for accessibility. The user might add CSS styling to improve accessibility. So whether publishers design for it or not is irrelevant.

CR has no DOM inspection so it does not seem important if some internal structures are added directly to the elements, rather than a separate structure index from the DOM, that would likely just be less efficient.

Just adding support to the text collector does not make the img alt text searchable. Perhaps give it a try as I did and you might then see the simplicity of organising the alt text as a child element, and ordered the ::before and ::after elements around as usual.

Frenzie:
> CR has no DOM inspection so it does not seem important if some internal structures are added directly to the elements, rather than a separate structure index from the DOM, that would likely just be less efficient.

We do. ;-)
Screenshot_20210113_170013

ourairquality:
For example, what do you do with the internal text elements in that inspector? I guess you map them to what is expected, so just do the same with the IMG element. Same with the pseudo elements, just ignore them in the inspector, or present them otherwise. Hiding internal data structures is trivial. I am just trying to point out that if the data can be organised to assist the walking of the text then that would be good, and a list of ordered child nodes is easy to walk, but perhaps there is another solution. Let me also have a look a the challenges for koreader, which seems very impressive btw, just no read-aloud support!

ourairquality:
Perhaps also have a look at the Chrome 'elements' inspector, it presents a linear list of child nodes which includes ::marker ::before ... ::after in the order that the content is presented, so it would seem ok for koreader to present these too, and it also presents ::before and ::after elements as child elements of the IMG element - but does not include the alt text as a child, but that seems fairly a trivial implementation dependent matter.

Frenzie:
> but that seems fairly a trivial implementation dependent matter.

Not quite that trivial. You'd break the HTML spec and thereby how you access the DOM through scripting. Which is quite irrelevant in the current CRE context, but trivial discrepancies like that between browsers could sometimes be quite the headache 15 years ago. I don't really have an opinion on the matter since I haven't looked at the code in question in sufficient detail recently (if ever), but it might be something to keep in mind.

ourairquality:
Closing. Would add that koreader goes to great lengths to present pdf documents in ways that the publisher did not intend, amazing really, and doing something similar for read-aloud does not appear to me to be a bug per sa, not just because the publisher did not intend that or because the publisher did not augment the content for accessibility.

Thanks for the feedback. I am going explore improving and augmenting the text walking code in CR, and have made some progress and a few dead ends already. If it become too complex or too limiting then perhaps HTML would be better translated into another linear structure for read-aloud, or just emitted as text first (with just the changes to the text collector it is possible to emit the entire document as text and read-aloud that back, simple), perhaps the DOM is not a suitable structure for that. Perhaps I will follow up on this when I have a coherent proposal and demo, rather than trying to upstream pieces.

poire-z:
Just came to mind that if you are the one adding CSS tricks to enhance accessibility and get spoken images, you can help yourself out of your hard problem of walking :before/alt/:after in order by just using one:
```css
img.blah::before { content: 'Start of image. ' attr(alt) ' End of image.'; }
```
<kbd>image</kbd>

No more hardcoded ugly trick with ALT as a text node, no need to worry about walking stuff in order, you decide the order via "content". You'd still have to get ldomNodeTextCollector to add the computed value of a pseudoElem to the accumulated text.

ourairquality:
Thank you, that is an interesting example to consider. If you try that in Chrome it repeats the alt text, so displays the alt text once in the ::before content and then repeats it after that. With this PR CR does the same as Chrome in this case and the CR Read-Aloud will repeat that too. Firefox does not appear to emit the ::before element here, not sure why, but does emit the ::after content and will also repeat the alt text in that case with the ::after content being after the alt text, and Chrome does the same in that case too, just as CR does with this patch. So the effective result of this PR on the presentation still appears valid even if the data structures are questionable.

WIP adds a new method like isText, but that returns true for leaf text nodes, so true for the internal text node and for pseudo elements. Also getText is trivially extended to return text from the pseudo elements. These simple changes improve a lot of code paths. e.g this CSS content becomes searchable in the walker, works with the text and word collectors, works mapping positions back to these nodes for selection etc.

There are lots of other issues to work on, so perhaps lets take a break from this PR.

e.g. The code walking still needs a lot of work in general and needs to be more robust if the pseudo elements can be walked for their content, and it helps Read-Aloud if it can find sentences cleanly - working on that.

e.g. If the list markers could be moved to ::marker pseudo elements than that would then make them work with the above WIP too, and I have started looking into this but are lost on the rendering changes for that. This change could make styling of the markers work too and perhaps making CR follow DOM standard more closely, and would anyone be interested in helping with this if I put up a WIP PR or made a start?

ourairquality:
Perhaps also interesting is that the Firefox inspector reports the img alt text as being a ::before element, even if no ::before element has been defined, so at a guess Firefox uses hard coded ::before element to implement the img alt attribute here, and that would also explain why it does not emit user supplied ::before content here.

poire-z:
> If the list markers could be moved to ::marker pseudo elements than that would then make them work with the above WIP too, and I have started looking into this but are lost on the rendering changes for that.
This change could make styling of the markers work too and perhaps making CR follow DOM standard more closely, and would anyone be interested in helping with this if I put up a WIP PR or made a start?

I have mixed feelings about that.

Initial one:
Well, I'd rather prefer we don't do that. For regular list items that can be many, we don't want to add a pseudoElem for each of them - and the CSS stuff involved that will surely slow rendering. We also don't support counter() stuff in content:, and that would not be easy to implement (I looked at it when doing content:, we can't easily have a state for counters when applying CSS).
We have list items code that works quite ok, and was complicated to get right. Re-doing all that mostly to get spoken list items seems... excessive.
In the places where you probably added handling of pseudoElem (_getText is trivially extended to return text from the pseudo elements_), you could just as well trivially extend it to IMG and LI in the same way: get_appplied_content_property could be just get_as_non_visual_text() for them, that would return the ALT value for IMG, and a computed "1 2 bullet 4" for LI by walking/counting its parents.
No real need to go at reworking the DOM and visual rendering.

Second one:
Adapting visual rendering for LI/::marker could be an interesting challenge: instead of being handled out of DOM with getMarker(&marker_width), we would get them in a pseudoElem and sized with getRenderedWidths(), initNodeRendMethod would make the node display:block if list-item-position:outside but display:inline or inline-block when :inside (would need to handle padding/margin for inline elements, which we don't, and is another complicated thing for the text rendering code) , the pseudoElem node would be erm_final/erm_block and have a RenderRectAccessor with x/y/w/h so it's fully positionned and sized in its container instead of being handled a bit too specifically in the rendering and drawing code...
It's really an interesting project, that touches many things - that would probably require me (who is confortable with all the things involved) a couple of weeks of work.

Third one:
Although interesting, I have other things on my todo list (stuff that don't yet work, and need to be implemented) and I don't really want to take on this project (re-doing stuff that work, for the sake of fun and doing it a bit more properly) - or even driving/reviewing/hinting/negotiating you if you decide to take it on :)

ourairquality:
I've submitted a few patches that reduce the diffs and that might be picked up. I have it somewhat working, the changes were not great, it does not yet handle counter content, just 'normal' or content already supported. It might not be restrictive enough on the styling, e.g. it accepts a background color, probably other such nits. Yes it's not going to be a big issue for Read-Aloud if you don't want to add the ::marker element. I'll put up a WIP soon, but will make another pass considering your comments, thanks. I got the pseudo element to be erm_final but it would still not map back from a pointer to the node, noticed a fmt height being zero which looked wrong - shall study the RenderRectAccessor and how to position and size it in its container.

PR #226, opened 2021-01-12
HTML: add a decimal point suffix to the ordered lists markets

ourairquality:
Web browsers and at least some other ereaders seem to emit a decimal point as a suffix to ordered list markers, so follow that. This can also help scanning the text because the points clearly mark a sentence end.

### DIFF
--- a/crengine/src/lvtinydom.cpp
+++ b/crengine/src/lvtinydom.cpp
@@ -17657,6 +17657,7 @@ bool ldomNode::getNodeListMarker( int & counterValue, lString32 & marker, int & 
         switch (st) {
         case css_lst_decimal:
             marker = lString32::itoa(counterValue);
+            marker << '.';
             break;
         case css_lst_lower_roman:
             if (counterValue > 0 &&
@@ -17664,6 +17665,7 @@ bool ldomNode::getNodeListMarker( int & counterValue, lString32 & marker, int & 
                 marker = lString32(lower_roman[counterValue-1]);
             else
                 marker = lString32::itoa(counterValue); // fallback to simple counter
+            marker << '.';
             break;
         case css_lst_upper_roman:
             if (counterValue > 0 &&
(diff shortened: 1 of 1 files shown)


poire-z:
First impression: I prefer without it (what are those misaligned dust grains? :)

image image

image image

But I dunno, I may get used to it.

Might be english thing? @Frenzie ?
I see these periods in footnotes in the few French paper books I just looked at - but they are all "list-style-position: inside" - where I think it may indeed "help scanning the test". Not so obvious when "list-style-position: outside" like on my screenshots.

ourairquality:
The common web browsers and Calibre viewer and other at least some other ebook readers do emit this decimal point. I did notice that CR right justifies the markers whereas the web browsers left justify them and that aligns the decimal points and might be better presentation - but that seems a separate issue to explore.

poire-z:
> CR left justifies the markers whereas the web browsers right justify them and that aligns the decimal points and might be better presentation 

Quick text with just (not sure if that's just enough, would need to check all combinations of list-style-position/text-align/rtl - and possibly tweak the space added to marker_width but not marker in ldomNode::getNodeListMarker(), probably depending on the combinations):
```diff
--- a/crengine/src/lvrend.cpp
+++ b/crengine/src/lvrend.cpp
@@ -8713,5 +8713,5 @@ void DrawDocument( LVDrawBuf & drawbuf, ldomNode * enode, int x0, int y0, int dx
                     LFormattedTextRef txform( enode->getDocument()->createFormattedText() );
                     // If RTL, have the marker aligned to the right inside list_marker_width
-                    lUInt32 txt_flags = is_rtl ? LTEXT_ALIGN_RIGHT : 0;
+                    lUInt32 txt_flags = is_rtl ? LTEXT_ALIGN_LEFT: LTEXT_ALIGN_RIGHT;
                     int list_marker_width;
                     lString32 marker = renderListItemMarker( enode, list_marker_width, txform.get(), -1, txt_flags);
@@ -8888,5 +8888,5 @@ void DrawDocument( LVDrawBuf & drawbuf, ldomNode * enode, int x0, int y0, int dx

                     // If RTL, have the marker aligned to the right inside list_marker_width
-                    lUInt32 txt_flags = is_rtl ? LTEXT_ALIGN_RIGHT : 0;
+                    lUInt32 txt_flags = is_rtl ? LTEXT_ALIGN_LEFT : LTEXT_ALIGN_RIGHT;
                     int list_marker_width;
                     lString32 marker = renderListItemMarker( enode, list_marker_width, txform.get(), -1, txt_flags);
```
And it would look rather fine:

image -- image


But may be not so fine when displayed as footnotes:

image -- image

Frenzie:
> Might be english thing? @Frenzie ?

I think most publications opt for a dot but I'd have to double check. I think the spec is more relevant though. :-)

For general info, counter styles:
https://drafts.csswg.org/css-counter-styles-3/#extending-css2

The relevant here, suffix (initial value of `. `):
https://drafts.csswg.org/css-counter-styles-3/#counter-style-suffix

ourairquality:
The links were useful, thanks. The standard defines a space in the suffix too, and perhaps we should follow that and make adjustments elsewhere if necessary, but not right now as there are too many patches pending in this area. They also defined the square as 25FE, it is a little larger than what is proposed elsewhere, but perhaps we should use that - added it to the comments and the replacement characters at least for now.

PR #227, opened 2021-01-13
HTML ordered lists: correct 'reversed' attribute handling

ourairquality:
The list should be reversed if the 'reversed' attribute is present,
irrespective of the value of the attritube.

### DIFF
--- a/crengine/src/lvtinydom.cpp
+++ b/crengine/src/lvtinydom.cpp
@@ -17568,7 +17568,7 @@ bool ldomNode::getNodeListMarker( int & counterValue, lString32 & marker, int & 
             ldomNode * parent = getUnboxedParent();
 
             // See if parent has a 'reversed' attribute.
-            int increment = parent->getAttributeValue(attr_reversed).empty() ? +1 : -1;
+            int increment = parent->hasAttribute(attr_reversed) ? -1 : +1;
 
             // If the caller passes in a non-zero counter then it is assumed
             // have been already calculated and have the value of the prior


virxkane:
https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#boolean-attribute

> If the attribute is present, its value must either be the empty string or a value that is an ASCII case-insensitive match for the attribute's canonical name, with no leading or trailing whitespace.
> The values "true" and "false" are not allowed on boolean attributes.

But it doesn't say how to interpret invalid values.
Also, main browsers ignore them, so I think everything is fine here.

Frenzie:
> But it doesn't say how to interpret invalid values.

But it does. It says there that the attribute's value should be either empty or the name of the attribute's canonical name itself. That explicitly overrides the more general rule that the mere presence or absence of the attribute itself is sufficient, but restricts it to `<ol reversed>`, `<ol reversed="">` and `<ol reversed="reversed">`.

To be clear, I'm not arguing we follow the spec here, but it sounds clear cut to me.

virxkane:
> But it does. It says there that the attribute's value should be either empty or the name of the attribute's canonical name itself.

Yes, of course, I usually read what I link to :). But that html spec doesn't say how to interpret invalid values. I.e if the value is invalid, there is no explicit indication to treat it as true or false.

Frenzie:
If that phrase about the value weren't there, it'd be true because the attribute is there. But because that phrase is present, it's false, because it explicitly limits the applicability of the general rule.

virxkane:
> If that phrase about the value weren't there, it'd be true because the attribute is there. But because that phrase is present, it's false, because it explicitly limits the applicability of the general rule.

I do not understand what you disagree with?
With this PR?
Or with my comment? If so, show link to the standard where this situation is explicitly described (how to interpret the invalid value of a boolean attribute).

Frenzie:
With your comment. The PR is exactly as it should be. But it looks like you're right, albeit in a sense I disagree with, and my memory was wrong. The alternative interpretation is indeed possible, but it's nevertheless nonsensical in the larger context.

## HTML 4

<https://www.w3.org/TR/PR-html40-971107/convent.html>
> If a user agent encounters an attribute it does not recognize, it should ignore the entire attribute specification (i.e., the attribute and its value). 
> If a user agent encounters an attribute value it doesn't recognize, it should use the default attribute value.

The default attribute value is that the entire attribute is *absent*. One can play robot or devil's advocate and say the default value is the attribute name, but if so why even bother mentioning valid values at all?

<https://www.w3.org/TR/PR-html40-971107/intro/sgmltut.html#h-3.3.4.2>
> Boolean attributes may legally take a single value: the name of the attribute itself (e.g., selected="selected").

## XHTML 1.0
<https://www.w3.org/TR/xhtml1/#uaconf>
> If a user agent encounters an attribute value it does not recognize, it must use the default attribute value.

According to the DTD we're not talking about XML boolean attributes here, are we? Because that'd sure simplify matters. Then `true`/`false` would be valid, but the default value would be `false`, which would make the behavior completely free from loopholes and stupid robots.

## XHTML 1.1

Refers back to the above, see <https://www.w3.org/TR/xhtml-modularization/conformance.html#s_conform_user_agent>. Although curiously it omits that specific point about the default attribute value, only including the point preceding it:

> If a user agent encounters an attribute it does not recognize, it MUST ignore the entire attribute specification (i.e., the attribute and its value).

## HTML 5
> 3.2.4.1 Attributes
> An attribute value is a string. Except where otherwise specified, attribute values on HTML elements may be any string value, including the empty string, and there is no restriction on what text can be specified in such attribute values.

<https://html.spec.whatwg.org/#restrictions-on-content-models-and-on-attribute-values>
> For example, setting the disabled attribute to the value "false" is disallowed, because despite the appearance of meaning that the element is enabled, it in fact means that the element is disabled (what matters for implementations is the presence of the attribute, not its value).

Curiously, HTML 5 takes the HTML 4 situation a step further by specifying the parsing instructions to death, without being fully explicit about how to deal with these disallowed restrictions. So it goes.

ourairquality:
There might be old documents that have a 'reversed' attribute with no value. Checked Chrome and it interprets that as a reversed ordering. That seems enough to make this trivial change.

Frenzie:
The PR is highly desired. I do not wish to follow the spec.

PR #229, opened 2021-01-15
CSS 'content': parse 'normal' separately from 'none'

ourairquality:
The 'normal' value has a separate meaning for the ::marker content,
it is not the same a 'none' for the ::before and ::after content.

### DIFF
--- a/crengine/src/lvstsheet.cpp
+++ b/crengine/src/lvstsheet.cpp
@@ -1075,7 +1075,8 @@ bool parse_content_property( const char * & str, lString32 & parsed_content)
     // consisting of a first letter, indicating its type, and if some
     // data: its length (+1 to avoid NULLs in strings) and that data.
     // parsed_content may contain multiple values, in the format
-    //   'X' for 'none' (or 'normal', = none with pseudo elements)
+    //   'X' for 'none'
+    //   'A' for 'normal'
     //   's' + <len> + string32 (string content) for ""
     //   'a' + <len> + string32 (attribute name) for attr()
     //   'Q' for 'open-quote'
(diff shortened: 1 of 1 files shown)


poire-z:
Oh my...

#229, #230, #231 should all be part of a same PR (the first 3 commits in it).

- There's no reason for us (CR/KR) to review intermediate bits and accept/merge some of them (just because they are no-op and cause no harm, but bring nothing until later) before having a chance to see where the whole thing is going. Keep these intermediates on your own tree until they all make some sense.
- It will be a nightmare for your as your work advances as many PRs will be conflicting.

(As for me personally: if you're going to work on markers, I might be looking at your progress, but I'm not commiting on taking that journey with you and spending as much time as I already did these last few days on your other PRs to pinpoint/explain all the things you'll have to think about and take care of - more than https://github.com/buggins/coolreader/pull/225#issuecomment-760772479 .)

ourairquality:
Ok, thought it would help to minimize the diffs. Some reviews like changes split up.

poire-z:
Well, your 4 PRs would be fine as intermediate commits in a single branch/PR.
Once you're done (or at big enough milestones and needing feedback), you can squash/rebase your work bits into commits that make sense individually - it will allow the reviewer to review individual units, while seeing where the whole thing is going (for example: https://github.com/koreader/crengine/pull/299/commits).

Frenzie:
> Ok, thought it would help to minimize the diffs. Some reviews like changes split up.

Depends on the changes. I can't speak for this repo of course, but as @poire-z said I think this looks like it should be a number of commits within the same PR, so you can easily see both the whole it builds to and the individual commits. A PR with a single commit of basically nothing like #231 comes across as a bit random. ^_^

PR #232, opened 2021-01-15
list marker: separate the content, max width, and rendering functions

ourairquality:
renderListItemMarker() had been also computing the maximum width,
walking all the list elements to do so if the value was not cached,
and this part has been split out into a separate function
listItemMarkerMaxWidth() which can simply obtain the cached value if
that is all the caller needs.

getNodeListMarker() had been also computing a width for the marker
content, and that computation has been split out into the few callers,
this leaves this function returning the content of the marker and
brings the code closer to supporting list ::marker content.

Basically a suggested clean up of this code, if you think that it is cleaner separating these function. This might support slightly better performance in cases where the caller did not need all the computation performed.

Could not spot any functional reasons that these function were combining the width computation. Only thought is that if the computation of the maximum width could fail somehow then it could fallback to the width of the one element, but that does not appear to be an issue.

Perhaps there was considered to be a possible need to adjust the width in some cases, where this would not just be based just on the content, but such hacks would not work with ::marker user defined content.

### DIFF
--- a/crengine/src/lvrend.cpp
+++ b/crengine/src/lvrend.cpp
@@ -2470,39 +2470,16 @@ void SplitLines( const lString32 & str, lString32Collection & lines )
 // marker_width is updated and can be used to add indent or padding necessary to make
 // room for the marker (what and how to do it depending of list-style_position (inside/outside)
 // is left to the caller)
-lString32 renderListItemMarker( ldomNode * enode, int & marker_width, LFormattedText * txform, int line_h, lUInt32 flags ) {
+lString32 renderListItemMarker( ldomNode * enode, LFormattedText * txform, int line_h, lUInt32 flags ) {
     lString32 marker;
-    marker_width = 0;
     ldomDocument* doc = enode->getDocument();
-    // The UL > LI parent-child chain may have had some of our boxing elements inserted
-    ldomNode * parent = enode->getUnboxedParent();
-    ListNumberingPropsRef listProps =  doc->getNodeNumberingProps( parent->getDataIndex() );
-    if ( listProps.isNull() ) { // no previously cached info: compute and cache it
-        // Scan all our siblings to know the widest marker width
-        int counterValue = 0;
-        int maxWidth = 0;
-        ldomNode * sibling = parent->getUnboxedFirstChild(true);
-        while ( sibling ) {
-            lString32 marker;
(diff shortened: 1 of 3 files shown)


crengine/src/lvrend.cpp
```diff
@@ -3067,17 +3078,22 @@ void renderFinalBlock( ldomNode * enode, LFormattedText * txform, RenderRectAcce
...
-                    if ( sibling->getNodeListMarker( counterValue, marker, markerWidth ) ) {
-                        if ( markerWidth > maxWidth )
-                            maxWidth = markerWidth;
+                    if ( sibling->getNodeListMarker( counterValue, marker ) ) {
+                        LVFontRef font = sibling->getFont();
+                        if ( !font.isNull() ) {
+                            TextLangCfg * lang_cfg = TextLangMan::getTextLangCfg(sibling);
+                            int markerWidth = font->getTextWidth(marker.c_str(), marker.length(), lang_cfg) + font->getSize()/8;
+                            if ( markerWidth > maxWidth )
+                                maxWidth = markerWidth;
```
poire-z, line 3087: May be this (from a few lines above what I reference) could go thru listItemMarkerMaxWidth() ? as it seems we do the same kind of work.

poire-z (commented):
I'd be fine with this cleanup (these list items bits are from very old times, I don't think anybody could tell about any functionnal reason to do it that way).
(PR closed, but just commenting on it to give some hints about what's expected before more work gets done.)

crengine/src/lvrend.cpp
ourairquality: Yes, after looking over that gain, thanks.

crengine/src/lvtinydom.cpp
```diff
@@ -17695,22 +17694,10 @@ bool ldomNode::getNodeListMarker( int & counterValue, lString32 & marker, int &
...
-        } else {
-            marker.clear();
-        }
-    }
-    return res;
 #else
     marker = cs32("*");
-    return true;
 #endif
+    return true;
```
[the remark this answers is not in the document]
ourairquality: It seem equivalent to the case of returning an empty marker, so the caller might as well just check for the empty marker?

If so then the return value could become the marker to make this even cleaner? All uses of this function pass in an empty marker as-is.

PR #236, opened 2021-01-21
preformatted text: don't remove trailing space from the last word on a line

ourairquality:
Noticed this issue exploring the ::marker support, where trailing spaces in content might be expected to remain. This can be reproduced in the following example so seems to be a more general issue - notice that the trailing space on the second line is short, and this seems unexpected and is not consistent with Chrome, so perhaps this issue and PR might be considered in isolation.

Also noticed that CR will format per-formatted text on multiple lines, where Chome will not - but that seems like a feature. With this change these lines retain their trailing space too, not just that last line. This works well when used for a marker content as it offsets the marker nicely even if it wraps on multiple lines.

```
<pre style="background: grey; text-align: right;">123456789</pre>
<pre style="background: grey; text-align: right;">123456   </pre>
```

### DIFF
--- a/crengine/src/lvtextfm.cpp
+++ b/crengine/src/lvtextfm.cpp
@@ -3310,7 +3310,10 @@ public:
                         word->min_width = word->width;
                         word->flags |= LTEXT_WORD_CAN_HYPH_BREAK_LINE_AFTER;
                     }
-                    if ( m_flags[i-1] & LCHAR_IS_SPACE) { // Current word ends with a space
+
+                    bool preformatted = srcline->flags & LTEXT_FLAG_PREFORMATTED;
+                    if ( m_flags[i-1] & LCHAR_IS_SPACE ) {
+                        // Current word ends with a space.
                         // Each word ending with a space (except in some conditions) can
                         // have its width reduced by a fraction of this space width or
                         // increased if needed (for text justification), so actually
@@ -3326,8 +3329,11 @@ public:
                                 word->min_width = word->width - dw;
                             }
                         }
-                        if ( lastWord ) {
-                            // If last word of line, remove any trailing space from word's width
+                        if ( lastWord && !preformatted ) {
+                            // If last word of line, remove any trailing space
+                            // from word's width (but not with preformatted, in
+                            // case of text-align:right where we don't want to
+                            // lose any trailing space)
                             word->width = m_widths[i>1 ? i-2 : 0] - (wstart>0 ? m_widths[wstart-1] : 0);
                             word->min_width = word->width;
                         }


poire-z:
> where trailing spaces in content might be expected to remain

How so?
I fail to see where this PR change would apply with preformatted: this bit of code is supposed to update a word width/min-width mostly used only when applying text alignment, and is useful with text justification, which then won't have a space at the end of last word, so last word fit the right margin.
I don't see how not doing that can be useful with preformatted text (except may be with text selection? or may be when you try to text-align: right such a marker and you want to keep the space?)
(The rendered width to estimate the best width for a paragraph of words (a final block) is handled elsewhere - if that trailing space should be kept in in that case.)

About markers, I saw in https://developer.mozilla.org/en-US/docs/Web/CSS/::marker that not many CSS properties apply to them - only inline properties, and none of the display properties. So, contrary to what I may have suggested at https://github.com/buggins/coolreader/pull/225#issuecomment-760772479 - these pseudoElem markers would probably best not be final blocks. I think a good way to handle them would be to just set them erm_invisible (so they are not another node to position and draw), and have the code that size or draw markers just pick the font color/size from the node and the text from its content: , in the same code that deal with normal list item markers (so not much of that code would need to be changed).

ourairquality:
Please just consider the example supplied and the formatting error demonstrated by that example:

Chrome:
```
....123456789
....123456___
```

CR3QT (the trailing space has been trimmed):
```
....123456789
.....123456__
```

The patch is intended to fix this, it stops the trimming of that trailing space in pre-formatted text.

This does not impact text selection, it is in the rendering path.

It is a struggle for me learning the code, and this may well not be the appropriate fix for this issue, but it does fix the example above. Perhaps someone who understands these code paths well could consider the issue and make a suggestion.

This appears to be a separate issue from supporting ::marker, and the approach to implementing ::marker (if at all) does not eliminate this issue. But, yes it was noticed affecting the presentation of right aligned list markers in some experimental code, trimming a space from the right aligned markers. It is possible to compensate for that in 'normal' markers by adding more space etc, but with user supplied ::marker content, where the user adds trailing space to affect the presentation, this becomes an issue there. Chrome does not trim that space when presenting the ::marker content.

Btw the current CR rending code for list markers does not right justify the markers, it's not so pretty left aligned particularly for numerical markers with a decimal point suffix. Improving this is not absolutely trivial, the current code either left aligns or it switches to using 'inside' list item presentation, see  ```renderAsListStylePositionInside()```, but please lets not get distracted by that, I can submit a WIP PR for that experimental code when it is half ready if you wish, and it makes an attempt to properly align the markers without the hack of switching to 'inside' list presentation, and it appears to work with RTL content too. My attempt to add support for ::marker may well be a dead end, but a lot of paths are working now as a proof of concept.

poire-z:
> Please just consider the example supplied and the formatting error demonstrated by that example:

OK, sorry, I missed the text-align:right (I hardly ever thought about PRE aligned right... I always think about it mostly as real `<pre>`, so left aligned).

> this may well not be the appropriate fix for this issue, but it does fix the example above. Perhaps someone who understands these code paths well could consider the issue and make a suggestion.

Well, no, it feels like the right fix. May be just add a comment so we remember why:
```diff
 // If last word of line, remove any trailing space from word's width
+// (but not with preformatted, in case of text-align:right where we don't want to lose any trailing space)
```

PR #240, opened 2021-01-23
legacy list item: remove

ourairquality:
There does not appear to be any practical reason to keep the legacy list item code paths now.

The legacy list item code path appears to have used a tab and a negative margin offset to achieve the presentation, with a hack in the text formatting code path for this.

If there is a good reason to keep this legacy code path around then please let me know, but it does not appear to be necessary and frustrates code development.

Was it just due to a lack of confidence in new code paths, so that there was a fallback, and is there confidence now and can these legacy code paths be remove now?

Was it to maintain pixel perfect presentation following legacy expectations?

If there is another reason then perhaps that could be documented, and perhaps there is a different way to achieve that requirement in future development?

### DIFF
--- a/crengine/include/cssdef.h
+++ b/crengine/include/cssdef.h
@@ -31,7 +31,6 @@ enum css_display_t {
     css_d_inline,  // All elements starts being css_d_inline, unless otherwise specified in fb2def.h
     // Above are those that define a mostly inline container, below those that define a mostly block container
     css_d_block,
-    css_d_list_item_legacy, // display: -cr-list-item-final (was used before 20180524 for display: list-item)
     css_d_list_item_block,  // display: list-item
     css_d_inline_block,
     css_d_inline_table, // (needs to be before css_d_table, as we use tests like if: (style->display > css_d_table))
--- a/crengine/src/lvrend.cpp
+++ b/crengine/src/lvrend.cpp
@@ -2523,12 +2523,7 @@ lString32 renderListItemMarker( ldomNode * enode, int & marker_width, LFormatted
                 flags |= LTEXT_STRUT_CONFINED;
         }
         marker += "\t";
-        // That "\t" had some purpose in css_d_list_item_legacy rendering to mark the end
-        // of the marker, and by providing the marker_width as negative indent, so that
-        // the following text can have some constant indent by rendering it just like
-        // negative/hanging text-indent. It has no real use if we provide a 0-indent
-        // like we do below.
-        // But coincidentally, this "\t" acts for fribidi as a text segment separator (SS)
+        // This "\t" acts for fribidi as a text segment separator (SS)
         // which will bidi-isolate the marker from the followup text, and will ensure,
         // for example, that:
         //   <li style="list-style-type: lower-roman; list-style-type: inside">Some text</li>
@@ -2791,8 +2786,7 @@ void renderFinalBlock( ldomNode * enode, LFormattedText * txform, RenderRectAcce
         }
 
         if ( (flags & LTEXT_FLAG_NEWLINE) && ( rm == erm_final || ( legacy_rendering && is_block ) ) ) {
-            // Top and single 'final' node (unless in the degenerate case
-            // of obsolete css_d_list_item_legacy):
+            // Top and single 'final' node:
             // Get text-indent and line-height that will apply to the full final block
             // There is also an exception: in legacy rendering mode, we must also indent any blocks.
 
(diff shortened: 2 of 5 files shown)


poire-z:
There is one on the KOReader side (dunno how this affects CoolReader):
Books opened in 2015 which have bookmarks/highlights have the xpointers of these highlights stored - and the dom version (or the absence of it as it did not exist at the time).
These xpointers could be no more valid if rendered with a newer dom version - or with using css_list_item_block which makes things block instead of _legacy which made them mostly inline - which could cause (I think, haven't checked, don't want to check) some autoBoxing elements to happen or not in the xpointer/xpath.
If that book is opened in 2021, we need to be able to load that book with the exact DOM it had in 2015 to be able to keep the highlights shown. And since late 2019, we can migrate them to the latest DOM version - but we need to open it once with the old DOM to do that migration.
(This might feel like a why bother about past users, things evolve, too bad for them. But I have cared about that over the last years. I may have broken them, I dunno :) but I made my best efforts not to.)

My take on crengine development is to limit any refactoring or even uneeded indentation/reformating: I need git blame to be of value to understand what happened/from when and why this line is from.
I personally don't care much about making it clean and easy - it's all very complex and every line or checks has some history. So, in my position and for my own devel bits, knowing now quite a bit, I value history more than accessibility. By going at refactoring/simplifying, you may easily lose edge cases and fixes. It's fragile, be careful, don't play too much with it.
(I don't want to make a nice and pretty libcrengine - I just want it to be good enough and stable so I can use it for reading - and don't refix what has been fixed.)

ourairquality:
I am not seeing an incompatibility in the paths, so I don't think this affected the highlights or bookmarks.

For example here is a path into a list item from CR 3.2.2-1: ```/body/DocFragment[2]/body/ol[10]/li[2]/text().1```, and this is also the same format as used in the current code. Changing the rendering paths for the list item has not affect this one.

But when adding ::marker and inserting a new child element some fixup is needed, but that is not this PR and might be a dead end.

> My take on crengine development is to limit any refactoring or even uneeded indentation/reformating.

This PR is not a refactoring, it removes code paths, it does not add codes paths, it does not rewrite code, it should only decrease the burden on maintenance.

> I personally don't care much about making it clean and easy - it's all very complex and every line or checks has some history.

> (I don't want to make a nice and pretty libcrengine - I just want it to be good enough and stable so I can use it for reading - and don't refix what has been fixed.)

Do you speak for the project being in maintenance mode?

If the code is basically in maintenance mode (or some parts of the code such as the crengine) then please just let people know so you are not bothered with development suggestions on it, or perhaps so that people keep suggestions within scope of what might be considered?

Perhaps I am misguided. Would anyone wish to opinion on where open source ereaders might be heading? It appears that it just needs a good xml parser and very little extra code to get something basic working, that a basic ereader is not so challenging if presentation expectations are low. I see some open source ereaders using mupdf as the presentation backend and mupdf even has epub support, but while open source the development seems rather closed. I see web engines being used as backends for presentation but they are rather heavy weight - they have very active development and testing. CRengine appears to have a strong focus on text presentation quality and it appears to do a relatively good job at presenting HTML and it appears to be much lighter weight than the web engines, but if it has all become too complex to support ongoing development then is it becoming a dead end, will open source ereaders inevitably move to either use web engines or alternatively keep their presentation ambitions modest?

virxkane:
crengine, especially lvtinydom develops thanks to @poire-z, I would listen to his opinion.
The idea of refactoring or simplifying the engine code is not without sense, but the code is quite complex, heavy and fragile, it is easy to break it and you will notice it only after a couple of months, when some book suddenly stopped displaying correctly. This has happened a number of times, so I would be very careful about these files: lvtextfm, lvtinydom, lvrend, lvstsheet, lvxml.
Unfortunately, we do not have a test suite system for engine, so it is impossible to unequivocally understand what harmless editing of one file will lead to.

> Do you speak for the project being in maintenance mode?

Not certainly in that way. New features are added only if the effect obtained exceeds the risks of breaking something in the engine.
@poire-z, correct me if I was wrong somewhere.

Frenzie:
@ourairquality 
> Do you speak for the project being in maintenance mode?

I don't think that reflects the current reality at all, unless you define maintenance mode as not pulling the rug out from under users constantly. :-)

What @poire-z describes above may have been in part my suggestion back around 2016 or 2017. For some context, please take a look at the commits in this repo we call upstream from ~2015-2018: https://github.com/buggins/coolreader/compare/cr3.1.2-71...cr3.2.2-1

Around that time upstream wasn't really active or taking patches, and we saw an unfortunate divergence/separate development as a result. You can get a general idea of it by looking at something like https://github.com/koreader/crengine/pulls?q=is%3Apr+is%3Aclosed+created%3A2016-01-01..2019-01-01+

I've personally never gotten too involved with the actual code — I've read & tested plenty of it and offered an extra pair of eyes, but when @poire-z really started getting into improving crengine I suggested to keep changes as minimal as reasonably possible to make it easy to compare with upstream and/or other forks. Besides which, we're all quite fond of git blame and formatting changes "break" a quick git blame. @virxkane was able to integrate the vast majority of it in this repo when it resumed activity, so I like to think that paid off. The past couple of years have been very nice, with improvements going in both directions (one of my favorites being the addition of HarfBuzz).

But to make a long story short, along the way @poire-z has also put in a fair bit of effort not to break things for users with old highlights and bookmarks. That's the reason for the legacy code path. As the comment says, `// obsolete (used only when gDOMVersionRequested < 20180524)`. Perhaps it's overly conservative in this case, but that is the general purpose of such sections. The possible lack of confidence you mention in your OP would more accurately be called a lack of confidence in the new code not breaking old bookmarks, although I suspect it more likely to reflect an actual known incompatibility..

> I see some open source ereaders using mupdf as the presentation backend and mupdf even has epub support

If you have a KOReader build available, I added (very basic) support for opening FB2 and EPUB with MuPDF a couple of years ago by long-pressing on the document and choosing "Open with". I might also look into their new chapter-based API.

> will open source ereaders inevitably move to either use web engines

I'm not sure to what extent that's possible. crengine is effectively closer to Prince than to a typical web engine. What I mean by that is the focus is on pages, while for web renderers pages are more of an afterthought. But I do think it's likely for performance to… let's say inevitably ;-)… become less of a concern.

@virxkane 
> Unfortunately, we do not have a test suite system for engine, so it is impossible to unequivocally understand what harmless editing of one file will lead to.

That's something I brainstormed about a few years ago. I never did quite get around to it because the W3C test suites didn't quite fit with what we'd need.

Taking a test like <http://test.csswg.org/suites/selectors-3_dev/nightly-unstable/html/root-siblings.htm>, the simplest method would be to take a screenshot and to compare that. But then improvements in unrelated matters like strutting, kerning, perhaps also slightly changed default margins, etc. would result in a fail.

```html
<!DOCTYPE html>
<html><head><meta charset="utf-8">
<title>CSS Test: Parsing check for * ~ :root error handling</title>
<link href="http://www.microsoft.com/" rel="author" title="Microsoft">
<link href="https://drafts.csswg.org/selectors-3/#selector-syntax" rel="help">
<link href="reference/ref-this-text-should-be-green.htm" rel="match">
<meta content="* ~ :root should not match anything" name="assert">
<style>
* ~ :root p {
  color: red;
}
p {
  color: green;
}
</style>
</head><body><p>This text should be green.</p>
</body></html>
```

So it seemed more fruitful to me to write the comparison in the form of a simpler method to achieve the same thing. That way, we can say the expected rendering looks this:

```html
<!DOCTYPE html>
<html><head><meta charset="utf-8">
<title>CSS Test: Parsing check for * ~ :root error handling</title>
<style>
p {
  color: green;
}
</style>
</head><body><p>This text should be green.</p>
</body></html>
```

But I never quite got past the brainstorming stages. :-)

virxkane:
> @virxkane was able to integrate the vast majority of it in this repo when it resumed activity

This work was started by @pkb, for which he is a special thankfulness.

NiLuJe:
Any other not-vaporware modern take on this (besides Plato) would probably invovle WebKit in one way or another (e.g., [fread](https://github.com/fread-ink/fread.ui), which is WIP and on hiatus, and I have no idea how far it went, just throwing it out there).

Special mention for https://github.com/joeycastillo/The-Open-Book, although with a different scope ;).

Frenzie:
I'd suggest you could also do pretty interesting things quite easily with Prince, if it ran on arm anyway. (I'm a bit of a fan, in case you're wondering.)

But yeah, WebKit does some good stuff. It's not necessarily that heavy either, if you don't feed it overly heavy, complex documents.

ourairquality:
Thank you for all the feedback and suggestions, and I now better understand what patches might be considered and the reasons.

CRengine appears to me to have a place, and perhaps it's underrated by many, but my guess is that it's place might be in being light weight and with a code base that a small number of people can manage and command.

So it looks like the lack of regression testing might be making it difficult to work on some areas of the code. I am not up the writing a regression testing framework, and presentation conformance is not my interest, but I might try and put together epub file(s) that test specific code paths that I work on to at least allow a manual re-test, and lets see if that helps at all. I did find a few similar attempts online:

http://epubtest.org/test-books
https://github.com/kobolabs/reading-platform-tests/tree/master/epub
https://github.com/IDPF/epub-testsuite/tree/master/content/parked-tests/epub30-test-0104/EPUB/xhtml

It is looking easy to adapt CRending for better Read-Aloud, and that is something, thanks.

PR #241, opened 2021-01-23
Fallback fonts management

virxkane:
* Implemented management over the list of fallback fonts in Qt & Android build. Related to #218.
* Added security check in fallback font processing: potential infinite recursion.
* Fixed incorrect value of "left_hyphen_min" and "right_hyphen_min" values, as well as the "index out of range" bug.

settings-new-qt-m
settings-new-android-s

### DIFF
--- a/android/jni/Android.mk
+++ b/android/jni/Android.mk
@@ -49,8 +49,8 @@ CRENGINE_SRC_FILES := \
     ../../crengine/src/cp_stats.cpp \
     ../../crengine/src/lvstring.cpp \
     ../../crengine/src/lvstring8collection.cpp \
-    ../../crengine/src/lvstring16collection.cpp \
-    ../../crengine/src/lvstring16hashedcollection.cpp \
+    ../../crengine/src/lvstring32collection.cpp \
+    ../../crengine/src/lvstring32hashedcollection.cpp \
     ../../crengine/src/crlog.cpp \
     ../../crengine/src/serialbuf.cpp \
     ../../crengine/src/props.cpp \
--- a/android/res/values-bg/strings.xml
+++ b/android/res/values-bg/strings.xml
@@ -381,6 +381,8 @@
     <string name="options_font_hinting_disabled">Изключено</string>
     <string name="options_font_hinting_auto">Автохинтинг на шрифта</string>
     <string name="options_font_fallback_face">Резервен шрифт</string>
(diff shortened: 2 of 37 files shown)


ourairquality:
Gave it a try, cr3qt and android, and it seems to work really well, thanks.

Noticed:

* Appears that the ```options_font_fallback_face``` strings could be removed
* Might be nice to have ```options_font_fallback_faces``` strings for all locales.
* Perhaps remove all remnants of ```SetFallbackFace```, lvfntman.h, lvfreetypefontman.h, lvdocview.cpp, 
* And remnants of ```PROP_FALLBACK_FONT_FACE```, lvdocview.cpp, lvdocviewprops.h, 
* There is still a reference to  ```PROP_FALLBACK_FONT_FACE``` in cr3gui/src/settings.cpp and is there an easy workaround to append a default set of fonts and update ```PROP_FALLBACK_FONT_FACES```, or could it just be commented out temporarily.

virxkane:
@ourairquality Thank you for your comment.

> * Perhaps remove all remnants of `SetFallbackFace`, lvfntman.h, lvfreetypefontman.h, lvdocview.cpp,
> * And remnants of `PROP_FALLBACK_FONT_FACE`, lvdocview.cpp, lvdocviewprops.h,

Ok, I can remove this.

> * Appears that the `options_font_fallback_face` strings could be removed
> * Might be nice to have `options_font_fallback_faces` strings for all locales.

Unfortunately, I do not speak all these languages. But remaining line `options_font_fallback_face` may help in translating the string `options_font_fallback_faces`. Although, really, I will try.


> * There is still a reference to  `PROP_FALLBACK_FONT_FACE` in cr3gui/src/settings.cpp and is there an easy workaround to append a default set of fonts and update `PROP_FALLBACK_FONT_FACES`, or could it just be commented out temporarily.

I think that this part of the source code is no longer relevant.

virxkane:
I did not delete the "options_font_fallback_face" string because I could not translate the "options_font_fallback_faces" and "options_font_fallback_face_num" strings into all languages, let it be a sample.

ourairquality:
Perhaps just use the ```options_font_fallback_face``` strings for ```options_font_fallback_face```, then they remain and are close, and someone else need just convert them into plural form. Could leave a TODO comment in the source so that people know that some corrections are expected to be needed.

PR #242, opened 2021-01-23
Fallback fonts: rework the list management

ourairquality:
Add support for storing a default fallback font face list which is
appended when setting a single font face.

The front ends are only setting a single fallback font, but also rely
on a larger fallback font sets which is either the default crengine
list or a platform specific list. This was failing because the code
paths had been setting a single user specified fallback font and then
overriding this with the fallback list.

The code could also get stuck with a filtered list of fallback fonts
and did not retry the filter in the case of new fonts being added. The
new default list is not filtered and when the fallback font changes
they are all retried which now picks up new fonts that might be
available.

This was my re-attempt at getting this working, but I see a much my extensive attempt at https://github.com/buggins/coolreader/pull/241 which I will look at, but here are some ideas too.


virxkane:
What do you dislike about the #241?
Or do you think you shouldn't give the user the ability to customize fallback fonts?

ourairquality:
Had not seen your work while working on this. Allowing the user to customize the fallback fonts list look great so shall look over that, thanks.

PR #243, opened 2021-01-24
Fribidi 1.0.10

ourairquality:
An attempt to update fribidi to the 1.0.10 release, tested on android. Also pulled in two upstream fixes that appear to be safe and important, an issue with bidi isolation and a pointer issue found in fuzzing.

I am chasing what appears to be a failure of bidi isolation, and there have been a good few fixes in fribidi in this area since 1.0.8, and there seem to be little point chasing known and fixed issues - this does not fix all the issues I see though, but they may well be my own programming errors.

This could use some independent testing and confirmation that it works on android. Perhaps if someone reading challenging RTL texts and who could spot regressions quickly - that's not me.


buggins:
covered with PR#244

PR #244, opened 2021-01-25
Third party libraries managements

virxkane:
Most of the sources of third-party libraries have been removed from the repository, and instead a script "thirdparty-deploy.sh" has been added that downloads them and patches them if necessary.
Information about the libraries managed by the script "thirdparty-deplay.sh" is located in the "thirdparty_repo" directory, the library sources are deployed in the "thirdparty" directory, and now the version is present in the name of the library directory. Therefore, when updating any library, you need to update not only the `<pkgname>.meta.sh` file, but also the `CMakeLists.txt`, `android/app/CMakeLists.txt`, `android/jni/Android.mk`, `android/app/thirdparty_libs/*` project files.
Library sources that are not managed by the script are now located in the "thirdparty_unman" directory. These are mainly libraries that are no longer being developed and libraries that are not planned to be updated.


Frenzie:
Incidentally, also see
https://cmake.org/cmake/help/git-stage/module/FetchContent.html (which is
slightly harder to use, but possibly more cross-platform).

NB Not a suggestion/complaint/whatever, just an fyi. ^_^

virxkane:
@Frenzie I was thinking about the git submodule, cmake ExternalProject, but did not find an acceptable solution for how to reapply the patches when updating any library.

PR #245, opened 2021-01-26
Add option to ignore document internal margins

ourairquality:
Some epub documents set margins that can become a problem viewing on
small screen devices, so add an option to ignore these.

btw KOreader appears to have an extensible feature for adding style mods and a similar one for this. The PR is a minimal change to be consistent with the existing CR code.

Tested on epub and html documents.

### DIFF
--- a/android/res/raw/chm.css
+++ b/android/res/raw/chm.css
@@ -1,6 +1,9 @@
 head, style, form, script { display: none; } 
 
-body { $def.all } 
+body {
+    $def.all
+    $body.margin
+}
 
 DocFragment { page-break-before: always }
 
--- a/android/res/raw/epub.css
+++ b/android/res/raw/epub.css
@@ -1,7 +1,8 @@
 body {
     text-indent: 0;
-    margin: 0;
     text-align: justify;
+    margin: 0;
+    $body.margin
 }
 
 head, script {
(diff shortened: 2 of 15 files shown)


virxkane:
Didn't see any effect from the new option in the desktop version.

virxkane:
Ok, on Desktop (Qt5) it works. Would it be more logical to move this option to the "stylesheet" tab?

ourairquality:
> Ok, on Desktop (Qt5) it works. Would it be more logical to move this option to the "stylesheet" tab?

Thank you for testing.

A key distinguishing character of this option is that it overrides the internal document style, whereas the other style options are additions that can have effect even if the internal document styles are disable and they do not necessarily override the internal style.

Could have added a 'body' stylesheet option (actually that is what I tried at first), and checkboxes to make them optionally important, but it seemed like overdoing it.

There are other 'margins' settings, on the 'page' table, but they are not style based, rather appear to apply a margin to the rending engine and can still be used even with the document internal page margins disabled. It did not appear to fit well there (tried that too!).

So it is suggested to be placed near the option to enable the document internal style, and it is disabled when these are disabled, so that seem a good placement to make that relationship as obvious as possible.

Perhaps someone will add more general style hints extensions in future making it obsolete, like in koreader, but it seems a very useful option to have for now because as-is some epub documents are practically unreadable with the internal styles enabled. E.g. they specify 10's of em in margins and on a small screen the content areas diminishes to nothing.

PR #246, opened 2021-01-27
CSS: add units vw vh vmin vmax

ourairquality:
These units should be useful for epub documents. They might also complement the min/max-width/height work being done on the koreader crengine where a percentage unit for these might not be usable, and a trivial variant of this PR was also tested on koreader.

Had to get at the page width and height in ```lengthToPx()``` and chose to pass around a node as there is always a node handy, but a document might be passed in but there is not always a document at hand. Passing in the page width and height would appear to be a burden for the callers, but if that is what people would prefer then that is also possible.

### DIFF
--- a/crengine/src/lvrend.cpp
+++ b/crengine/src/lvrend.cpp
@@ -452,8 +452,7 @@ public:
                             col->percent = w.value / 256;
                         }
                         else if ( w.type != css_val_unspecified ) { // px, em...
-                            int em = item->getFont()->getSize();
-                            col->width = lengthToPx( w, 0, em );
+                            col->width = lengthToPx( item, w, 0 );
                             // (0 as the base width for %, as % was dealt with just above)
                         }
                         // otherwise cell->percent and cell->width stay at 0
@@ -536,8 +535,7 @@ public:
                             cell->percent = w.value / 256;
                         }
                         else if ( w.type != css_val_unspecified ) { // px, em...
-                            int em = item->getFont()->getSize();
-                            cell->width = lengthToPx( w, 0, em );
+                            cell->width = lengthToPx( item, w, 0 );
                         }
                         // else: cell->percent and cell->width stay at 0
 
@@ -855,9 +853,8 @@ public:
         }
 
         css_style_ref_t table_style = elem->getStyle();
-        int table_em = elem->getFont()->getSize();
         // border-spacing does not accept values in % unit
-        int borderspacing_h = lengthToPx(table_style->border_spacing[0], 0, table_em);
+        int borderspacing_h = lengthToPx(elem, table_style->border_spacing[0], 0 );
         bool border_collapse = (table_style->border_collapse == css_border_collapse);
 
         if (border_collapse) {
@@ -1105,8 +1102,8 @@ public:
         }
         else { // no collapse
             // Remove table outer paddings (margin and padding indexes are LRTB)
-            assignable_width -= lengthToPx(table_style->padding[0], table_width, table_em);
-            assignable_width -= lengthToPx(table_style->padding[1], table_width, table_em);
+            assignable_width -= lengthToPx(elem, table_style->padding[0], table_width);
+            assignable_width -= lengthToPx(elem, table_style->padding[1], table_width);
             // Remove (nb cols + 1) border-spacing
             assignable_width -= (cols.length() + 1) * borderspacing_h;
         }
@@ -1292,11 +1289,11 @@ public:
             // distribute restw to columns, but shrink table width
             // Table padding may be in %, and need to be corrected
             int correction = 0;
-            correction += lengthToPx(table_style->padding[0], table_width, table_em);
-            correction += lengthToPx(table_style->padding[0], table_width, table_em);
+            correction += lengthToPx(elem, table_style->padding[0], table_width);
+            correction += lengthToPx(elem, table_style->padding[0], table_width);
             table_width -= restw;
-            correction -= lengthToPx(table_style->padding[0], table_width, table_em);
-            correction -= lengthToPx(table_style->padding[0], table_width, table_em);
+            correction -= lengthToPx(elem, table_style->padding[0], table_width);
+            correction -= lengthToPx(elem, table_style->padding[0], table_width);
             table_width -= correction;
             #ifdef DEBUG_TABLE_RENDERING
                 assignable_width -= restw + correction; // (for debug printf() below)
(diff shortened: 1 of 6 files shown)


Frenzie (approved):
Fwiw, this is how I would've done it.

crengine/src/lvrend.cpp
```diff
@@ -453,7 +453,7 @@ class CCRTable {
                         }
                         else if ( w.type != css_val_unspecified ) { // px, em...
                             int em = item->getFont()->getSize();
-                            col->width = lengthToPx( w, 0, em );
+                            col->width = lengthToPx( el, w, 0, em );
```
poire-z, line 456: Use `item` (the child we're dealing with) instead of `el` (the parent).

crengine/src/lvrend.cpp
```diff
@@ -1469,10 +1469,10 @@ class CCRTable {
             fmt.setWidth( w ); // fmt.width must be set before 'caption->renderFinalBlock'
                                // to have text-indent in % not mess up at render time
             css_style_ref_t caption_style = caption->getStyle();
-            int padding_left = lengthToPx( caption_style->padding[0], w, em ) + measureBorder(caption, 3);
-            int padding_right = lengthToPx( caption_style->padding[1], w, em ) + measureBorder(caption,1);
-            int padding_top = lengthToPx( caption_style->padding[2], w, em ) + measureBorder(caption,0);
-            int padding_bottom = lengthToPx( caption_style->padding[3], w, em ) + measureBorder(caption,2);
+            int padding_left = lengthToPx( elem, caption_style->padding[0], w, em ) + measureBorder(caption, 3);
```
poire-z, line 1472: `caption` instead of `elem` (the table). Same below, should be `cell->elem`.

crengine/src/lvrend.cpp
```diff
@@ -2433,6 +2433,28 @@ int lengthToPx( css_length_t val, int base_px, int base_em, bool unspecified_as_
         if (gRenderDPI)
             px = 96 * value / 6 >> 8;
         break;
+    case css_val_vw: {
+        int page_width = node->getDocument()->getPageWidth();
+        px = val.value * page_width / 25600;
```
poire-z, line 2438: May be make that `/ 100 >> 8` like the other ones.

crengine/src/lvrend.cpp
```diff
@@ -3897,15 +3919,15 @@ int pagebreakhelper(ldomNode *enode,int width)
...
+    int padding_top = lengthToPx( enode, enode->getStyle()->padding[2], width, em ) + DEBUG_TREE_DRAW+measureBorder(enode,0);
     flag=CssPageBreak2Flags(getPageBreakBefore(enode))<<RN_SPLIT_BEFORE;
     if (flag==RN_SPLIT_BEFORE_ALWAYS){
         ldomNode *node=enode;
         int top=0;
         while (!node->isNull()) {
-            top+=lengthToPx( node->getStyle()->margin[2], width, em ) +
-                 lengthToPx( node->getStyle()->padding[2], width, em ) +
+            top+=lengthToPx( enode, node->getStyle()->margin[2], width, em ) +
+                 lengthToPx( enode, node->getStyle()->padding[2], width, em ) +
```
poire-z, line 3930: `node` (the parent we are iterating) instead of `enode` (the child we started from).
(Haven't checked all, just throwing this one that was obvious from the diff.)

ourairquality:
Thank you for the feedback, and indeed most of the base_em values passed to lengthToPx() could be obtained from the node now, but not all. So made it an optional argument and eliminated the obvious which was most of them, updating the nodes passed to be the nodes originally used to obtain the base_em.

Moving gRootFontSize to the document sounds like a good idea now, but lets do that separately.

Did not adjust any of the 'tweaks' yet, might leave that for follow up work, but these new units appeared useful in CR as-is, tested on some text boxes and images. But yes these units should be widely useful, not need disabling.

KOreader appeared to need more work on the 'tweaks', some examples tried did not respond, but some did. It seems to ignore max-height which might not be necessary with these units. I can send them a PR too.

crengine/src/lvrend.cpp
```diff
@@ -3518,7 +3536,7 @@ void renderFinalBlock( ldomNode * enode, LFormattedText * txform, RenderRectAcce
                     << baseflags << ")\n";
             #endif
 
-            ldomNode * parent = enode->getParentNode();
+            ldomNode * const parent = enode->getParentNode();
```
poire-z, line 3539: Just curious (I'm an amateur at C++):
- is `ldomNode * const parent` exactly the same as `const ldomNode * parent` ? If yes, please put the `const` at start, like I'm used to see it elsewhere (and you did it above in `const int em`) so there's no doubt :)
- how/why do you decide to add `const` ? Is this a reflex ? Do you have to look aroud to be sure it's not updated ? Or do you add it by default and see if the compiler complains?

poire-z:
> KOreader appeared to need more work on the 'tweaks', some examples tried did not respond, but some did. It seems to ignore max-height which might not be necessary with these units.

Yes, best to test with *-width stuff. The handling of height is a bit limited, see https://github.com/koreader/crengine/pull/409 (but it should be ok on images, where only height in % are not ensured.)
(Mostly because we fit height to the content. We can allow a height/min-height when it is larger than the content, but we never allow the block height to be smaller than its content. Mostly because of the way we add "slices" of content to make "pages" (which web browsers don't have to care about), and we don't want to go back on the Y-axis and handle content overflowing their container in that process (and it's usually not really needed). (Although we might allow doing it for "small" things like negative margins.)

> I can send them (koreader) a PR too.

Well, that will avoid me doing the job of picking, but it won't help @virxkane job of picking the min/max-width stuff (that won't apply easily if this PR is merged here). So, "instead" would be better than "too".

> but these new units appeared useful in CR as-is

I don't doubt it :), and they will eventually be (with some delay) in CR. (But I understand it's frustrating to have to wait and work on various branches and have to wait on things that don't depend on you.)

ourairquality:
> > KOreader appeared to need more work on the 'tweaks', some examples tried did not respond, but some did. It seems to ignore max-height which might not be necessary with these units.
> 
> Yes, best to test with *-width stuff. The handling of height is a bit limited, see [koreader/crengine#409]

My mistake, koreader re-enabled the image scaling options and disabling them in the UI got my limited tests working - a feature!

> > I can send them (koreader) a PR too.
> 
> Well, that will avoid me doing the job of picking, but it won't help @virxkane job of picking the min/max-width stuff (that won't apply easily if this PR is merged here). So, "instead" would be better than "too".

CR should be able to pick up this PR without much issue. I have branches with these patches in either order and there are only two conflicting trivial functions. There would be a balancing benefit of the trees having less absolute differences.

Re: the const declarations, those were useful for double checking the changes in this PR. The compiler will check if the const variable has been reassigned when declared const. Most of them were removed in code that was removed.

There is a difference in the placement of the const. Above it means that the local variable reference (the stack allocation) does not change - these gave confidence that a node or em reference was not being reassigned. But it does not mean that the content of the object being referenced is constant and that might change if passed as an argument to another function - these can also be declared const in a more functional style of programming, so there might be multiple const declarations on the same line, a const reference to a const data object.

crengine/src/lvrend.cpp
NiLuJe: >     * is `ldomNode * const parent` exactly the same as `const ldomNode * parent` ? If yes, please put the `const` at start, like I'm used to see it elsewhere (and you did it above in `const int em`) so there's no doubt :)

It isn't, c.f., [the clockwise/spiral rule](http://c-faq.com/decl/spiral.anderson.html) ;).

(Here, the pointer itself is constant, not the data it points to).

ourairquality:
Best to pickup this one from koreader/crengine now, got a bit bigger in the end.

PR #247, opened 2021-01-28
CacheFileHeader: zero the alignment padding for reproducibility

ourairquality:
The CacheFileItem struct within the CacheFileHeader was being aligned by the
compiler to an 8 byte boundary creating a 4 byte whole in the header that was
not being zeroed and this was leading to unreproducible content in the cache
file. Add the padding explicitly and zero it.

### DIFF
--- a/crengine/src/lvtinydom.cpp
+++ b/crengine/src/lvtinydom.cpp
@@ -491,6 +491,9 @@ struct SimpleCacheFileHeader
 struct CacheFileHeader : public SimpleCacheFileHeader
 {
     lUInt32 _fsize;
+    // Padding to explicitly align the index block structure, and that can be
+    // be initialized to zero for reproducible file contents.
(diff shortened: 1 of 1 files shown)


virxkane:
sizeof(CacheFileHeader), before/after:
x86_64-pc-linux-gnu: 104/104
i686-linux-android: 100/104

virxkane:
https://software.intel.com/content/www/us/en/develop/blogs/understanding-x86-vs-arm-memory-alignment-on-android.html

On x86 this padding useless.

poire-z:
> it might not make sense to add any exceptions for x86

I agree, no need to bother with an exception, it's just 4 bytes more :)
(Similar attempt elsewhere with https://github.com/koreader/crengine/commit/a783b53 discuss at https://github.com/koreader/koreader/pull/4972#issuecomment-487123345)

Issue #250, opened 2021-01-31
Request supporting ttc font


jack6th:
Most of the system CJK fonts are ttc format.   
On Linux there are 'Noto San CJK' series and on Windows there are 'Simsun', 'MingLiu', 'Microsoft Yahei', ...   
Please support it.

jack6th:
Now the new version support ttc font.   
I installed Noto Sans CJK Regular and Noto Sans CJK Light on Windows.  
(1) I cannot see Light font in the dropdown list.
(2) If I change the CJK font in the new fallback fonts manager and press OK button, I cannot see the change result until I change the Text font and press OK button.

PR #251, opened 2021-02-01
initSystemFonts: recognise the 'ttc' font file type

ourairquality:
TrueType Collection (TTC) files are commonly distributed
e.g. Noto Sans CJK

Simply accepting the 'ttc' type files in initSystemFonts appears adequate to get these working in cr3qt on linux. The code appears to already be using the index.

https://github.com/buggins/coolreader/issues/250


ourairquality:
Simply recognising the 'ttc' font file type also appeared to work on Android. It might be necessary to have the new fallback font management UI to be able to add the CJK fonts, that is what I used to test this, https://github.com/buggins/coolreader/pull/241

PR #252, opened 2021-02-01
resizeImage: correct a scaling bug and rewrite for accuracy

ourairquality:
There was cases in which the scaled dimensions could be off by a factor of
1000. A scaling factor of 1000 had been used to likely improve accuracy, but
the calculations can be done more accurately otherwise. Rewrite the
calculations to fix the bug and to give accurate results.

### DIFF
--- a/crengine/include/cssdef.h
+++ b/crengine/include/cssdef.h
@@ -414,4 +414,27 @@ typedef struct css_length_tag {
     lUInt32 pack() { return (lUInt32)type + (((lUInt32)value)<<4); }
 } css_length_t;
 
+// Support
+
+// Integer division with rounding to nearest.
+//
+// This does not bother to 'round to nearest integer' in the case of a tie.
+static inline lInt32 roundi(lInt32 n, lInt32 divisor)
+{
+    if (divisor < 0) {
+        if (divisor == 0) {
+            if (n >= 0)
+                return 0x7fffffff;
+            return 0x80000000;
+        }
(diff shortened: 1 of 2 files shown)


virxkane:
@ourairquality Write how to test it. Some test case.

ourairquality:
I used a throwaway shadow function that used float calculations for testing, and compared the results of the two functions - thrown away. Then tried resizing pages of images visually looking for the correct scaling and checked the log for any difference between these integer calculations and the rounded float calculations and the only differences were for 0.5 fractions where the floats round up or down whereas the integer rounding here does not bother with that extra complexity as it seems unnecessary - but it is abstract to a function and could be easily implemented in future.

The initial problem with the 1000x dimension images was noted on Android, perhaps when selecting a very large font size (don't recall but it was unusable and a problem). It can be seen in the code by inspection, it is in the 'integer' scaling code path, notice scale_div is initialized to 1 and there are a number of code paths where it falls through to be used as a divisor as 1 where the numerator is always scaled by 1000, so there are a number of failure paths. Perhaps just initializing scale_div to 1000 would have fixed it but the code looks convoluted and inaccurate to me and it would be even more trouble to validate code that is known to be inaccurate.

ourairquality:
Moved the definition of ```roundi()``` into cssdef.h as it is useful across files, and also used by https://github.com/buggins/coolreader/pull/258 Also improved ```roundi()``` a little.

PR #255, opened 2021-02-02
android: fix DPI initialization

ourairquality:
The DPI was not being initialized correctly leaving it at the default
of 96 dpi, but the intention appears to have been to increase the CSS
DPI for higher DPI devices.

The SettingsManager() was being initialized early, before initializing
the display DPI, and was pushing the initial default DPI of 96 to the
crengine when calling loadSettings(). Move the initialization of the
display DPI earlier, into SettingsManager() to address this.

An alternative might have been to access the display DPI within loadSettings() just as the screen size is.

The default of 96 dpi was very small for a small display with a high dpi so this fix can be a big improvement. The display DPI obtained from the system might also vary based on system accessibility settings making CR a bit more responsive.

Btw: Would anyone want a 'zoom' UI setting, like in web browsers?  Took a look at doing this and a quick hack could just scale the DPI, but it might be better to add a separate zoom settings property to crengine as some web standards want separately the actual physical DPI and the zoomed DPI?

### DIFF
--- a/android/src/org/coolreader/crengine/BaseActivity.java
+++ b/android/src/org/coolreader/crengine/BaseActivity.java
@@ -102,6 +102,26 @@ public class BaseActivity extends Activity implements Settings {
 	private SettingsManager mSettingsManager;
 
 	protected void startServices() {
+		DisplayMetrics dm = new DisplayMetrics();
+		getWindowManager().getDefaultDisplay().getMetrics(dm);
+
+		try {
+			Field fld = dm.getClass().getField("densityDpi");
+			if (fld != null) {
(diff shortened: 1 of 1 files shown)


virxkane:
I find it wrong to put DPI definition code in BaseActivity.SettingsManager.SettingsManager(). `densityDpi` is declared in BaseActivity class, therefore, you need to write the value in any method of this class in order to prevent spaghetti code from growing. To solve the above problem, just move this code to the beginning of the BaseActivity.startServices() function.

ourairquality:
Moved to the start of BaseActivity.startServices().

PR #256, opened 2021-02-02
Update thirdparty-deploy.sh

pazos:
As it contains Bash-isms (or posix-isms) that are not compatible with some shells, like fish.

Any dev machine that has `/bin/sh` should also has `/bin/bash`. Since I don't see a bat alternative I don't expect to manage these repos from windows.

Pinging @virxkane

### DIFF
--- a/thirdparty-deploy.sh
+++ b/thirdparty-deploy.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 
 top_srcdir=`pwd`
 


virxkane:
@pazos ok, let's replace with '#/bin/bash'.
About Windows: I don't see the problem, we all use git, git for Windows contains 'git bash'. Naturally, this script was developed with possibility of running in Windows in 'git bash'.  And it works great in it.
But I certainly don't mind if someone writes an alternative in the form of a bat file.

PR #257, opened 2021-02-03
Adjust and add mimeType to support chm,fb2,fb3 for onyx

Edward259:


virxkane:
@Edward259 Thank you very much!
But one note about line `<data android:mimeType="application/octet-stream" />`:
with this line I can also open `*.sqlite` with CoolReader and may be other types too. Also on Google Nexus 5 `*.ttf` can be opened with CoolReader.
May be delete this line? Without it all declared file types already can be opened with CoolReader.

Edward259:
<data android:mimeType="application/octet-stream" />
About this, I just move it to mimeType position,or is will disturb fb2、fb3 open way.

Edward259:
`<data android:mimeType="application/octet-stream" />:`
About this, I just move it to mimeType position,or is will disturb fb2、fb3 open way.

PR #258, opened 2021-02-06
parse_number_value: accept '+' prefix and exponent

ourairquality:
CSS appears to allow a '+' prefix to numbers and also an exponent. The '+' is easy to handle. There is limited range for the exponent but this PR takes a positive exponent back from the power of 10 fraction dividend when possible so in these case it should work the same, e.g. 0.123456e6 is the same as 123456.

### DIFF
--- a/crengine/include/cssdef.h
+++ b/crengine/include/cssdef.h
@@ -414,4 +414,27 @@ typedef struct css_length_tag {
     lUInt32 pack() { return (lUInt32)type + (((lUInt32)value)<<4); }
 } css_length_t;
 
+// Support
+
+// Integer division with rounding to nearest.
+//
+// This does not bother to 'round to nearest integer' in the case of a tie.
+static inline lInt32 roundi(lInt32 n, lInt32 divisor)
+{
+    if (divisor < 0) {
+        if (divisor == 0) {
+            if (n >= 0)
+                return 0x7fffffff;
+            return 0x80000000;
+        }
+        n = -n;
+        divisor = -divisor;
+    }
+
+    if (n >= 0)
+        return (n + (divisor / 2)) / divisor;
+    else
+        return (n - (divisor / 2)) / divisor;
+}
+
 #endif // __CSS_DEF_H_INCLUDED__
--- a/crengine/src/lvstsheet.cpp
+++ b/crengine/src/lvstsheet.cpp
@@ -533,13 +533,16 @@ static bool parse_number_value( const char * & str, css_length_t & value,
             return true;
         }
     }
+
     bool negative = false;
-    if (accept_negative) {
-        if ( *str == '-' ) {
-            str++;
-            negative = true;
-        }
+    if (accept_negative && *str == '-') {
+        str++;
+        negative = true;
     }
+    else if (*str == '+') {
+        str++;
+    }
+
     int n = 0;
     if (*str != '.') {
         if (*str<'0' || *str>'9') {
@@ -551,6 +554,7 @@ static bool parse_number_value( const char * & str, css_length_t & value,
             str++;
         }
     }
+
     int frac = 0;
     int frac_div = 1;
     if (*str == '.') {
@@ -566,6 +570,29 @@ static bool parse_number_value( const char * & str, css_length_t & value,
             str++;
         }
     }
+
+    int exponent = 0;
+    bool exponent_negative = false;
+    if ((*str == 'e' || *str == 'E') &&
+        (str[1] == '+' || str[1] == '-' || (str[1] >= '0' && str[1] <= '9'))) {
+        str++;
+        if (*str == '-') {
+            str++;
+            exponent_negative = true;
+        }
+        else if (*str == '+') {
+            str++;
+        }
+        if (*str<'0' || *str>'9') {
+            str = orig_pos; // revert our possible str++
+            return false; // not a number
+        }
+        while (*str>='0' && *str<='9') {
+            exponent = exponent*10 + (*str - '0');
+            str++;
+        }
+    }
+
     if ( substr_icompare( "em", str ) )
         value.type = css_val_em;
     else if ( substr_icompare( "pt", str ) )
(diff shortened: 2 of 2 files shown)


poire-z:
Looks OK I guess. Pinging @NiLuJe for confirmation (because arithmetics, bits... negative bits arithmetics!).

> CSS appears to allow a '+' prefix to numbers and also an exponent

Please mention your references (to avoid us searching):
https://developer.mozilla.org/en-US/docs/Web/CSS/number
Not in CSS2.1: https://www.w3.org/TR/CSS2/syndata.html#value-def-number
but in CSS3: https://drafts.csswg.org/css-values-3/#numbers

crengine/include/cssdef.h
```diff
@@ -414,4 +414,27 @@ typedef struct css_length_tag {
     lUInt32 pack() { return (lUInt32)type + (((lUInt32)value)<<4); }
 } css_length_t;
 
+// Support
+
+// Integer division with rounding to nearest.
+//
+// This does not bother to 'round to nearest integer' in the case of a tie.
+static inline int roundi(int n, int divisor)
```
NiLuJe, line 422: @poire-z is not wrong, divisor ought to be an `unsigned int` to avoid potential UB.

crengine/include/cssdef.h
```diff
@@ -414,4 +414,27 @@ typedef struct css_length_tag {
...
+            if (n >= 0)
+                return 0x7fffffff;
+            return 0x80000000;
+        }
+        n = -n;
+        divisor = -divisor;
+    }
+
+    if (n >= 0)
+        return (n + (divisor >> 1)) / divisor;
```
NiLuJe, line 435: And then `divisor >> 1U` to make that very obvious ;).

crengine/include/cssdef.h
NiLuJe: Also, the C99 convention for those is to lead with the return type, so I'd rename this `iroundi` ;).

ourairquality:
Kept the divisor as a signed integer, as this function is intended to be a drop in replacement for signed division when working with the unit lengths which are signed. But did qualify the argument and result as being ```lInt32``` as that is the expected size that this function will be working with.

Replaced the bit shift with a division, to leave that up to the compiler to optimize, and to avoid the need to discuss that source of UB! This entire function depends on the number representation being two's complement so there seems no point trying to make it robust enough to work with any other number representation, but at least this is a move towards abstracting that into a few support functions.

Kept the name as ```roundi``` which parallels ```roundf```, that seems marginally better to me, but someone else is welcome to change that if they wish. There is more work to do here, this might not be a keeper, lets please not overthink this now.

I've a branch replacing the unit value with a float, and the time wasted on scrutinizing these types of bit operations is tilting me towards that. That branch keeps the screen px represented as an integer for now, with all other unit length types being a float, which might help performance after document conversion. Also have an branch using the current integer fixed point representation but falling back to intermediate float operations when there is a risk of overflow, to try to maintain the suspected integer performance on challenged systems. Let me do some profiling on low end systems to see if there is much in such choices, perhaps an RPi compiling for soft float. Perhaps people need to consider the trade offs including the developer burden.

poire-z:
> lets please not overthink this now

Second time you serve us this :/
Isn't "overthinking" what this and some recent PRs are about? Any real book example where handling +1.23e4 happens, and where a +1/-1 rounded pixel or +/- 1/ 256 of a length really matters?

ourairquality:
> > lets please not overthink this now
> 
> Second time you serve us this :/
> Isn't "overthinking" what this and some recent PRs are about? Any real book example where handling +1.23e4 happens, and where a +1/-1 rounded pixel or +/- 1/ 256 of a length really matters?

ok.

NiLuJe:
> Let me do some profiling on low end systems to see if there is much in such choices, perhaps an RPi compiling for soft float. Perhaps people need to consider the trade offs including the developer burden.

That sounds like a sensible approach :).

(FWIW, Kindle is soft-float, even on devices with neon (!). But we do target actual soft-float devices with a subpar VFP, too. Err, also, Kindles ;p).

poire-z:
Sorry if that sounded harsh. I have nothing against parsing 1.23e4 - even with rounding errors. It's the fight with rounding, and what it may start getting you thinking (putting floats in lvstyles), that worries me a bit.

Frenzie:
> Any real book example where handling +1.23e4 happens, and where a +1/-1 rounded pixel or +/- 1/ 256 of a length really matters?

While not directly relevant to this PR (since as you say… I'm not sure I've ever seen that used), you can easily come up with a case where a pixel more or less matters. Something like this for example:

```html
<!DOCTYPE html>
<title>Negative margin</title>
<style>
body {width:50vw; height:50vh; border:1px solid #000; padding: 0}
div {width:20px; height:20vh; border:1px solid #000; float:right; margin-top:-1px; margin-right:-22px}
</style>
<div></div>
```

Screenshot_2021-02-07_10-10-53

Make it off by 1px, and:

Screenshot_2021-02-07_10-11-23

ourairquality:
> Sorry if that sounded harsh. I have nothing against parsing 1.23e4 - even with rounding errors. It's the fight with rounding, and what it may start getting you thinking (putting floats in lvstyles), that worries me a bit.

Please just scavenge something from the PR if you wish, that might be more productive. This one seemed useful to consider as it makes CR a little more compatible and that might help the larger community by having a less brittle open source epub implementation - how you do that is not so important. I wish there were some process between filing an issue and submitting a PR, e.g. identifying an issue or enhancement with a suggested PR but not wanting to champion the PR.

Your work on crengine is a great contribution. I have been picking up patches before CR, thanks. I'll try to put up my own development branch in time and perhaps it will at least be food for thought or let people try some alternative, and perhaps you can scavenge something from there. It seems clear that the bulk of what I am doing will never make it back into the main repo, and much of it may well be a dead end, and I need time to explore that for myself, to understand the trade offs.

poire-z:
> I wish there were some process between filing an issue and submitting a PR, e.g. identifying an issue or enhancement with a suggested PR but not wanting to champion the PR.

Well, there's no defined process, but that's just happening with discussions here.
You can submit issues if you want some discussions before tackling it.
You can submit a PR if you have solved it - and accept the possible discussion and possibly rework it.
By now, you may have learned (the hard way) what we may be willing to accept and what can cause some friction and discussion. I was fine with this PR, until you mentionned pushing more things to become floats - which is ok to be discussed (and we should try to be more lenient with each other, me with your enthusiam, you with my coldness with heavy changes :)

> It seems clear that the bulk of what I am doing will never make it back into the main repo, and much of it may well be a dead end

PR #263, opened 2021-02-24
Fixes 2021.02-1

virxkane:
* Fixed infinity loop in `PageSplitState::SplitLineIfOverflowPage()`.
* Hides the onscreen software keyboard when opening the preferences dialog. TextEdit for editing the font size automatically opens this keyboard on some Android versions immediately upon opening a dialog.
* Removed useless screen refresh on e-ink devices.

### DIFF
--- a/crengine/src/lvpagesplitter.cpp
+++ b/crengine/src/lvpagesplitter.cpp
@@ -365,6 +365,10 @@ public:
     }
     void SplitLineIfOverflowPage( LVRendLineInfo * line )
     {
+        // If page_h <= 0 this function will fall into an infinite loop
+        // in which it will allocate memory until it runs out.
+        if (page_h <= 0)
+            return;
         // A 'line' is usually a line of text from a paragraph, but
         // can be an image, or a fully rendered table row (<tr>), and
         // can have a huge height, possibly overflowing page height.
--- a/android/res/layout/option_item_number.xml
+++ b/android/res/layout/option_item_number.xml
@@ -68,8 +68,9 @@
     </LinearLayout>
 
     <View
-            android:id="@+id/view"
+            android:id="@+id/margin_view"
             android:layout_width="4dp"
             android:layout_height="1dp"
-            android:layout_weight="0" />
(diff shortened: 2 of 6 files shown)


crengine/src/lvpagesplitter.cpp
```diff
@@ -365,6 +365,10 @@ struct PageSplitState {
     }
     void SplitLineIfOverflowPage( LVRendLineInfo * line )
     {
+        // If page_h <= 0 this function will fall into an infinite loop
+        // in which it will allocate memory until it runs out.
+        if (page_h <= 0)
+            return;
```
poire-z, line 371: Do you remember the context/book with which this happened ?
Because it looks to me page_h is fixed and comes from ldomDocument._page_height , which is set by setRenderProps(... dy...) so very early in LVDocView from the screen/page height.
(May be it happened with table cells, floats, inline-box, and I forgot to pass this page_height from an upper context?)

virxkane:
@poire-z I guess it never showed up for KOReader.
It's just that the CoolReader frontend for Android, upon initialization, first creates an LVDocView with a size of 100x100 and later sets the required size. Don't ask me why, I don't know and I'm not sure that I won't break anything if I try to change it :).
And if you set the top and bottom margins of 40 or more, page_h will be <= 0.

virxkane:
If we abstract and inspect the `PageSplitState::SplitLineIfOverflowPage()` function separately, we can clearly see that without this check it can get into an infinite loop.

poire-z:
> And if you set the top and bottom margins of 40 or more, page_h will be <= 0.

Somehow no issue on our side :) a normally 1 page document can become 2000 pages :)
image

It's just the whole code doesn't expect negative page height - so ok, it just infinite-loops here in this little function - but may be it would be safer to just force it to be `if page_h <= 0 then page_h=1` in the constructor - as there are probably other places where things could go bad even if no infinite loop.

Anyway, not minding about this edge case and your solution. Was just curious if this could happen in more normal circumstances :)

PR #271, opened 2021-03-21
Internal file manager background operations

virxkane:
* Added scanning progress indicator in the internal file manager.
* In the internal file manager, after changing the "hide directories without books" option in the settings, the directory is rescanned.
* Long directory scans have been moved to a background thread.
* The directory scan progress dialog is now cancelable.
* Scan directory progress dialog intervals updated.


Korb:
> * Added scanning progress indicator in the internal file manager.

In the current version, when you first launch the application and give it access to all smartphone files on the root screen, the "Browse file system" section is empty. No scanning indicators are displayed. I thought that the application was not working correctly, and now it can only be used to launch from other applications. I looked through all the settings, [the user manual](https://docs.google.com/document/d/1nJT_dwJpO7L-EJD7bBz7yShjebYK5zRxFbA5zKGjCyM/edit), but did not find any indication that, for example, this application can open documents only from a certain folder, but not from any on the device. The next day I reopened the app and the "SD", "EXT SD" and "Books" icons appeared in the "Browse file system" section.

I suggest making sure that the progress bar for scanning all files on the device is indeed displayed, regardless of the number of files on the device.

```
3.2.58-1-fdroid (32580)
Added 21.08.2021
Repo: F-Droid
Size: 33,3 MiB
Android: 1.6+
Requirements: arm64-v8a, armeabi-v7a, x86, x86_64
```

```
OS version: Funtouch OS 13 Global
Model: V2153
Hardware version: MP_0.1
Build number: PD2201OF_EX_A_13.1.13.8.W20.V000L1
Baseband version: MPSS.HA.1.1.c2-00012-KD_ALL_PACK-1.6347.75
Kernel version: 4.19.157-perf+ #1 Tue Jan 10 10:41:25 CST 2023
Compile time: 10 Jan 2023 05:28:48
Android security update: 1 January 2023
Google Play system update: 1 January 2023
```

PR #279, opened 2021-04-18
Fixes for 3.2.55 (tier 3)

virxkane:
* Don't call onCancel listener in 'check download setting' dialog if positive or negative button is pressed.
* Synchronization: deleting the lock file on failure or cancellation.
* Synchronizer related: fixed scanning of just downloaded books in zip-archive.
* InputDialog: the cursor position is retained while dragging the slider.

### DIFF
--- a/crengine/src/private/lvfreetypeface.cpp
+++ b/crengine/src/private/lvfreetypeface.cpp
@@ -350,7 +350,7 @@ LVFont *LVFreeTypeFace::getFallbackFont(lUInt32 fallbackPassMask) {
             index++;
             if (index == fallback_count)
                 index = 0;
-            _fallbackFont = fontMan->GetFallbackFont(_size, getWeight(), _italic, index);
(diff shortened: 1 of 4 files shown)


crengine/src/private/lvfreetypeface.cpp
```diff
@@ -1645,8 +1646,7 @@ lUInt16 LVFreeTypeFace::measureText(const lChar32 *text,
                 cur_width += letter_spacing_w;
             }
             widths[i] = cur_width;
-            if ( !isHyphen ) // avoid soft hyphens inside text string
-                prev_width = cur_width;
+            prev_width = cur_width;
```
poire-z, line 1649: Was this change voluntary ? Or is this taken care of elsewhere?
I think isHyphen maks soft-hyphens, that is a glyph that has a width. We don't draw them, so keeping their width in measureText() (as you're doing now here) might result in measureText() measuring words larger that what they will be when drawn.

virxkane: This is dead code.

PR #290, opened 2021-05-17
Android Studio project: f-droid support

### LINKED ISSUE #284: Request: a build flavor without GMS
Coolreader updates on F-droid are [blocked caused by the recent inclusion of Google services](https://gitlab.com/fdroid/fdroiddata/-/issues/2350). If I understand correctly they were added to support GDrive sync.

I would like to solve this but I'm not sure if upstream is wishing to spend time on reviewing this. Ideally it would be
a couple of folders under `android/app/src`. Lets call them `gplay` and `fdroid`. Within each folder a new class with the same package name and same class name.

That class would be the one that implements methods that rely on GMS within the program. This for the `gplay` flavor. The Fdroid flavor would include empty methods. Both classes should contain a boolean method called something like `isSupported` which returns true on the former and false on the later.

The changes needed on the common code would be checking that method response and invoke the specific method needed for a functionality. Something like:

```java
if (gms.isSupported()) {
    gms.doSomething(params)
}
```

And include gms as a depency only on  `gplayimplementation`

The other alternative would be to fork the project and remove GMS in the fork. That would be easier to implement but harder to keep updated. Pinging @virxkane

virxkane:
Android Studio project: created two flavors: main "primary" and "fdroid" for f-droid.
From "fdroid" flavor removed GSuite libraries and functionality for synchronization with Google Drive.
"Primary" flavor with no changes.
This should closes #284.


android/app/src/primary/AndroidManifest.xml
```diff
@@ -1,7 +1,10 @@
 <?xml version="1.0" encoding="utf-8"?>
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
         xmlns:tools="http://schemas.android.com/tools">
-        
+
+    <!-- Required permission to use in-app billing. -->
+    <uses-permission android:name="com.android.vending.BILLING" />
```
pazos, line 6: `tools:node="merge"` isn't required here?

virxkane: I'm not sure, but this is not necessary...
https://developer.android.com/studio/build/manifest-merge#node_markers

PR #299, opened 2021-06-11
Updates from koreader/crengine

virxkane:
Merged new stuff from koreader/crengine:
* merge https://github.com/koreader/crengine/pull/397
* merge part of https://github.com/koreader/crengine/pull/399 (and reworked)
* merge part of https://github.com/koreader/crengine/pull/402
* merge https://github.com/koreader/crengine/pull/405
* merge https://github.com/koreader/crengine/pull/406
* merge https://github.com/koreader/crengine/pull/409
* merge https://github.com/koreader/crengine/pull/413
* merge part of https://github.com/koreader/crengine/pull/416
* merge https://github.com/koreader/crengine/pull/419
* merge https://github.com/koreader/crengine/pull/421
* merge https://github.com/koreader/crengine/pull/423
* merge https://github.com/koreader/crengine/pull/425
* merge https://github.com/koreader/crengine/pull/428
* merge https://github.com/koreader/crengine/pull/429
* merge part of https://github.com/koreader/crengine/pull/430
* merge https://github.com/koreader/crengine/pull/431
* merge https://github.com/koreader/crengine/pull/436
* merge part of https://github.com/koreader/crengine/pull/438

Thirdparty library/dependencies added:
* zstd (for cache compression)
* utf8proc (improvements to unicode support).

Other changes:
* Android: optimized pixel format conversion from crengine BGRX to android RGBA. Conversion time was reduced by about 50%.
* Use the correct font glyph cache item size & glyph image offset. Based on https://github.com/koreader/crengine/pull/441
* Fixed global-buffer-overflow.
* Fixed coolreader/crengine/src/lvstring.cpp:127:42: runtime error: signed integer overflow: 1775760480 * 31 cannot be represented in type 'int'.
* lvtinydom.cpp: `bool hasInvisibleParent(ldomNode* node)`: null pointer dereference fixed.
* Memory leaks fixed.


crengine/src/lvtinydom.cpp
```diff
@@ -9056,8 +9056,14 @@ ldomXPointer ldomDocument::createXPointer( lvPoint pt, int direction, bool stric
                 }
 
                 lUInt32 hints = WORD_FLAGS_TO_FNT_FLAGS(word->flags);
-                font->measureText( str.c_str()+word->t.start, word->t.len, width, flg, word->width+50, '?',
-                            src->lang_cfg, src->letter_spacing + word->added_letter_spacing, false, hints);
+                if (str.empty() && word->t.len > 0) {
+                    // Don't know that the fuck up, but it happens
+                    for (int i = 0; i < word->t.len; i++)
+                        width[i] = 0;
```
poire-z, line 9062: Strange indeed.
Can you reproduce this? Specific book? Specific action that uses createXPointer() ?

crengine/src/lvtinydom.cpp
```diff
@@ -6406,9 +6406,11 @@ int initTableRendMethods( ldomNode * enode, int state )
 
 bool hasInvisibleParent( ldomNode * node )
 {
-    for ( ; !node->isRoot(); node = node->getParentNode() )
-        if ( node->getStyle()->display==css_d_none )
+    for ( ; node && !node->isRoot(); node = node->getParentNode() ) {
+        css_style_ref_t style = node->getStyle();
+        if (!style.isNull() && style->display == css_d_none)
```
poire-z, line 6406: Got a crash? Or just some reports from code analysis tools?
Because unless I missed something with the call from the added MathML bit, the only other call is from initNodeRendMethod, where there is always a style defined at this point.
And when walking nodes up, we should always meet the root node (so, !node->isRoot() will get us out of this loop), and so we should never meet a NULL node.

crengine/src/lvtinydom.cpp
```diff
@@ -7774,6 +7775,8 @@ void ldomNode::initNodeRendMethod()
             }
         }
     #endif
+
+    persist();
```
poire-z, line 7779: Asking, as you must have dig into that to understand it - and you added more in this and previous commit:
Is there a simple rule of thumb as to when we should use/add it?
No slowness added with doing it for each node here?

crengine/src/lvtinydom.cpp
```diff
@@ -4133,14 +4133,8 @@ ldomDocument::ldomDocument()
...
-    allocTinyElement(NULL, 0, 0);
-    // Note: valgrind reports (sometimes, when some document is opened or closed,
-    // with metadataOnly or not) a memory leak (64 bytes in 1 blocks are definitely
-    // lost), about this, created in allocTinyElement():
-    //    tinyElement * elem = new tinyElement(...)
-    // possibly because it's not anchored anywhere.
-    // Attempt at anchoring into a _nullNode, and calling ->detroy()
-    // in ~ldomDocument(), did not prevent this report, and caused other ones...
+    ldomNode* node = allocTinyElement(NULL, 0, 0);
+    node->persist();
```
poire-z, line 4137: I like when my questions get such simple answers ;)

crengine/src/lvtinydom.cpp
virxkane: > Can you reproduce this? Specific book?

Yes, of cource, book: http://fred-wang.github.io/mathml.css/

I don't know how to reproduce this in KOReader, but in the desktop version of CoolReader, just move the mouse over the text and the program will crash sooner or later.
Just see backtrace in commit comment.

crengine/src/lvtinydom.cpp
virxkane: > Got a crash?

Yes, of cource.


> Because unless I missed something with the call from the added MathML bit, the only other call is from initNodeRendMethod, where there is always a style defined at this point.

Crashed when style is null ref.

> And when walking nodes up, we should always meet the root node (so, !node->isRoot() will get us out of this loop), and so we should never meet a NULL node.

I just added it just in case, it costs absolutely nothing.

crengine/src/lvtinydom.cpp
virxkane: > Asking, as you must have dig into that to understand it - and you added more in this and previous commit:
> Is there a simple rule of thumb as to when we should use/add it?

I think yes. If you call `ldomNode::modify ()` to perform any node manipulation, then after that you must call `ldomNode::persist()`.

> No slowness added with doing it for each node here?

I don't know, not tested.

crengine/src/lvtinydom.cpp
virxkane: In `tinyNodeCollection::allocTinyElement()` we alloc new object, therefore, somewhere we have to delete it. `ldomNode::persist()` converts the mutable object to persistent and removes the `tinyElement` instance that would otherwise be lost.

crengine/src/lvtinydom.cpp
virxkane: Just look at the `ldomNode::setAttributeValue()` function where `ldomNode::modify()` is called. I think it would be silly to call `persist()` right there, it can be done later.

virxkane: See `ldomDocumentWriter::ElementCloseHandler()`.

Also you can try `tinyNodeCollection::persist()` instead of this.

crengine/src/lvtinydom.cpp
poire-z: Well, can't reproduce. Anyway, it feels that if this happens, the bug might be elsewhere, and it feels awkward handling that here.
If you have the time, could you try to printf the xpointer.toStringV1() when this happens, and try to find out what bit of HTML/MathML is involved in this document?

Just curious: why is there such a CR3View::mouseMoveEvent() handler ? Does it really get the xpointer at each mouse move ?! Feels it might feel really slow.

crengine/src/lvtinydom.cpp
poire-z: > Crashed when style is null ref.

Can you get a backtrace? I'd like to see from which of the 2 hasInvisibleParent() calls this happen.

crengine/src/lvtinydom.cpp
virxkane: > Just curious: why is there such a CR3View::mouseMoveEvent() handler ? Does it really get the xpointer at each mouse move ?! Feels it might feel really slow.

Sorry, I have zero interes to dig this depper.

crengine/src/lvtinydom.cpp
virxkane: > Can you get a backtrace? I'd like to see from which of the 2 hasInvisibleParent() calls this happen.

Sorry, I lost it. Document the same: http://fred-wang.github.io/mathml.css/

PR #308, opened 2021-08-30
Fixes for 3.2.58

virxkane:
* crengine: fixed integer overflow when measuring text with fallback fonts (HarfBuzz render).
* Android: When opening an external stream, find the corresponding file in the database to find the last read position and load the saved bookmarks, and if not found, save it to the application's private directory and save the information to the database.
* Fixed incorrect parsing of content file in 'chm' files.

Screenshot with integer overflow with HarfBuzz render:
device-2021-08-30-110202_sz

### DIFF
--- a/crengine/src/chmfmt.cpp
+++ b/crengine/src/chmfmt.cpp
@@ -11,6 +11,10 @@
 #include "../include/crlog.h"
 #include <chm_lib.h>
 
+#include "../include/fb2def.h"
+#define XS_IMPLEMENT_SCHEME 1
+#include "../include/fb2def.h"
+
 #define DUMP_CHM_DOC 0
 
 struct crChmExternalFileStream : public chmExternalFileStream {
@@ -866,10 +870,10 @@ ldomDocument * LVParseCHMHTMLStream( LVStreamRef stream, lString32 defEncodingNa
     if ( stream.isNull() )
         return NULL;
 
+#if 0
     // detect encondig
     stream->SetPos(0);
 
-#if 0
     ldomDocument * encDetectionDoc = LVParseHTMLStream( stream );
     int encoding = 0;
     if ( encDetectionDoc!=NULL ) {
@@ -905,11 +909,14 @@ ldomDocument * LVParseCHMHTMLStream( LVStreamRef stream, lString32 defEncodingNa
     ldomDocument * doc;
     doc = new ldomDocument();
     doc->setDocFlags( 0 );
+    doc->setNodeTypes(fb2_elem_table);
+    doc->setAttributeTypes(fb2_attr_table);
+    doc->setNameSpaceTypes(fb2_ns_table);
 
     ldomDocumentWriterFilter writerFilter(doc, false, HTML_AUTOCLOSE_TABLE);
     writerFilter.setFlags(writerFilter.getFlags() | TXTFLG_CONVERT_8BIT_ENTITY_ENCODING);
 
-    /// FB2 format
+    /// HTML format
     LVFileFormatParser * parser = new LVHTMLParser(stream, &writerFilter);
     if ( !defEncodingName.empty() )
         parser->SetCharset(defEncodingName.c_str());
@@ -1109,11 +1116,11 @@ public:
                 return false;
             }
 
-    #if DUMP_CHM_DOC==1
-        LVStreamRef out = LVOpenFileStream(U"/tmp/chm-toc.html", LVOM_WRITE);
-        if ( !out.isNull() )
-            doc->saveToStream( out, NULL, true );
-    #endif
+#if DUMP_CHM_DOC==1
+            LVStreamRef out = LVOpenFileStream(U"chm-toc.xml", LVOM_WRITE);
+            if ( !out.isNull() )
+                doc->saveToStream( out, NULL, true );
+#endif
 
             ldomNode * body = doc->getRootNode(); //doc->createXPointer(cs32("/html[1]/body[1]"));
             bool res = false;
--- a/crengine/src/chmfmt.cpp
+++ b/crengine/src/chmfmt.cpp
@@ -829,7 +829,7 @@ public:
     }
 
     lString32 getContentsFileName() {
-        if ( _binaryTOCURLTableId!=0 ) {
+        if ( _binaryTOCURLTableId!=0 && _urlTable != NULL ) {
             lString8 url = _urlTable->urlById(_binaryTOCURLTableId);
             if ( !url.empty() )
                 return decodeString(url);
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -7,7 +7,7 @@ buildscript {
         mavenCentral()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:7.0.0'
+        classpath 'com.android.tools.build:gradle:7.0.1'
         // NOTE: Do not place your application dependencies here; they belong
         // in the individual module build.gradle files
     }
--- a/android/jni/cr3engine.cpp
+++ b/android/jni/cr3engine.cpp
@@ -824,6 +824,32 @@ JNIEXPORT jbyteArray JNICALL Java_org_coolreader_crengine_Engine_scanBookCoverIn
 
 /*
  * Class:     org_coolreader_crengine_Engine
+ * Method:    isArchiveInternal
+ * Signature: (Ljava/lang/String;)Z
+ */
+JNIEXPORT jboolean JNICALL Java_org_coolreader_crengine_Engine_isArchiveInternal
+  (JNIEnv * _env, jclass, jstring jarcName)
+{
+    jboolean res = JNI_FALSE;
+    CRJNIEnv env(_env);
+    lString32 arcName = env.fromJavaString(jarcName);
+    LVStreamRef stream = LVOpenFileStream(arcName.c_str(), LVOM_READ);
+    if (!stream.isNull()) {
+        LVContainerRef arc = LVOpenArchieve(stream);
+        if (!arc.isNull()) {
+            if (!DetectEpubFormat(stream) &&
+                !DetectFb3Format(stream) &&
+                !DetectDocXFormat(stream) &&
+                !DetectOpenDocumentFormat(stream)) {
+                res = JNI_TRUE;
+            }
+        }
+    }
+    return res;
+}
+
+/*
+ * Class:     org_coolreader_crengine_Engine
  * Method:    getArchiveItemsInternal
  * Signature: (Ljava/lang/String;)[Ljava/lang/String;
  */
@@ -1179,6 +1205,7 @@ static JNINativeMethod sEngineMethods[] = {
   {"setCacheDirectoryInternal", "(Ljava/lang/String;I)Z", (void*)Java_org_coolreader_crengine_Engine_setCacheDirectoryInternal},
   {"scanBookPropertiesInternal", "(Lorg/coolreader/crengine/FileInfo;)Z", (void*)Java_org_coolreader_crengine_Engine_scanBookPropertiesInternal},
   {"updateFileCRC32Internal", "(Lorg/coolreader/crengine/FileInfo;)Z", (void*)Java_org_coolreader_crengine_Engine_updateFileCRC32Internal},
+  {"isArchiveInternal", "(Ljava/lang/String;)Z", (void*)Java_org_coolreader_crengine_Engine_isArchiveInternal},
   {"getArchiveItemsInternal", "(Ljava/lang/String;)[Ljava/lang/String;", (void*)Java_org_coolreader_crengine_Engine_getArchiveItemsInternal},
   {"isLink", "(Ljava/lang/String;)Ljava/lang/String;", (void*)Java_org_coolreader_crengine_Engine_isLink},
   {"suspendLongOperationInternal", "()V", (void*)Java_org_coolreader_crengine_Engine_suspendLongOperationInternal},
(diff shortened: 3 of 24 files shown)


crengine/src/lvtextfm.cpp
```diff
@@ -1717,7 +1717,7 @@ class LVFormatter {
         LVFont * lastFont = NULL;
         lInt16 lastLetterSpacing = 0;
         int start = 0;
-        int lastWidth = 0;
+        lUInt16 lastWidth = 0;
```
poire-z, line 1720: I think this one should stay `int`: we use int with `int m_widths`, which is the text formatter buffer and can be > 65535. It's only the local `widths` provided to lvfntman (which works only with lUInt16) that is limited to 65535

crengine/src/lvtextfm.cpp
```diff
@@ -1901,9 +1901,9 @@ class LVFormatter {
                         if ( lastSrc->flags & LTEXT_FLAG_PREFORMATTED )
                             scale_space_width = false;
                     }
-                    int cumulative_width_removed = 0;
-                    int prev_orig_measured_width = 0;
-                    int char_width = 0; // current single char width
+                    lUInt16 cumulative_width_removed = 0;
+                    lUInt16 prev_orig_measured_width = 0;
+                    lUInt16 char_width = 0; // current single char width
```
poire-z, line 1906: Does this fixes something ? Or you just made them lUInt16 because you know you were working with lUInt16 as returned by lvfntman's measureText() ?
It shouldn't hurt keeping them int, right ? As a int can't be overflowed when we give it lUInt16 ?
But I think `cumulative_width_removed` can get a chance to be negative (at line 1922), so better it stays `int` (or all stays `int` to take the less risks :)

crengine/src/lvtextfm.cpp
virxkane: We have complex spaghetti:
* `static lUInt16 widths[MAX_TEXT_CHUNK_SIZE+1]` in `LVFormatter::measureText()` (static local variable)
* `int * LVFormatter::m_widths`
* `lUInt16 formatted_word_t::width`
* few local variable of the type `int`
* `LVFont::virtual lUInt16 measureText(const lChar32 *text, int len, lUInt16 *widths, lUInt8 *flags, int max_width, lChar32 def_char, TextLangCfg * lang_cfg = NULL, int letter_spacing = 0, bool allow_hyphenation = true, lUInt32 hints = 0, lUInt32 fallbackPassMask = 0)`

And it's all intertwined together. Most likely, I did not fully understand all this. But I mean it is not the size that is important here, but the sign. Sign-reversal integer overflow is painful. I thought I need to get rid of the potential problem. Although not completely fixed :)
Do you really think that `lastWidth` should be of type int?

crengine/src/lvtextfm.cpp
virxkane: To be honest, I am not ready to investigate all of this, but if you are sure that this change is doing more harm, I am ready to revert it.

crengine/src/lvtextfm.cpp
virxkane: Ok, reverted. There is no desire to study these thousands of lines :)

poire-z: > Do you really think that lastWidth should be of type int?

Yes, int or unsigned int or lInt32 or lUInt32.

> We have complex spaghetti

Not my cooking :) but well, it makes sense to me:

We may do multiple runs of lastFont->measureText() because it's indeed limited to measuring up to what can fit in a lUInt16 (and we provided a max_width=0x7FFF, so half of it, for safety may be, or to avoid signed overflow in case we use some signed local variables without thinking much ?)
Then, after each run, with lastWidth, we always do just:
`lastWidth += widths[len-1]` , that is appending what we measured in the run - and
`m_widths[start + k] = lastWidth + widths[k]` as m_widths can exceed 0xFFFF (fortunately, we can measure paragraphs that are longer than 65535 pixels wide).
Feels like with `unsigned int`, we could measure twice (0xFFFFFFFF) than with `int` (0x7FFFFFFF), but well, they would both be super very long paragraphs, so we should be fine with either.

> lUInt16 formatted_word_t::width

I guess 65535 px is fine for a single word :)

crengine/src/lvtextfm.cpp
poire-z: I'm not sure it does more harm, but it feels it can :)
I would let lvtextvm.cpp unchanged.

> But I mean it is not the size that is important here, but the sign

Yes, but if we get lUInt16, we can easily store/add them to a `int` - the `int` will stay positive, and shouldn't hurt if we put/add it back in a lUInt16 (for individual ones) or accumulate them (in which case a `unsigned int` would allow more, but we're fine with just a `int` if we don't want to spend much time thinking about it and possible side effects).

But I think your other changes in lvfreetypeface.cpp are right (hard to get back into all that :).
Btw, can you share the document from your screenshot (or part of it)? I didn't manage to reproduce the bug that this should solve, even by making a super long paragraph in a fb2 - might depends on the font setups, dunno...

virxkane: I think paragraph with 2000 characters long will be enough with font size about 80 px.
Set main font without characters from main book language (I use "Android Clock"), set HarfBuzz render and try. On my side the first fallback font is also not suitable ("Noto Color Emoji"), but I think it has no effect.

[Chadovich_Tropa_1_Evangelie-ot-Timofeya.vNFIxg.167219.fb2.zip](https://github.com/buggins/coolreader/files/7076209/Chadovich_Tropa_1_Evangelie-ot-Timofeya.vNFIxg.167219.fb2.zip)
( Scroll to last page )

poire-z: Thanks, I could reproduce.
I think your fix is fine - but I'm not really familiar with the lastFit handling.

(There is this:
```diff
             // i is used below to "fill props for rest of chars", so make it accurate
-            i = len; // actually make it do nothing
+            i = lastFitChar;
```
My memory sucks, and I don't remember if I had a good reason for this "actually make it do nothing" or if I was just too lazy understand why filling flags for next ignored chars was needed - but I also have the vague memory it was needed in some other context, but dunno if that was before Harfbuzz cluster rewrite or after...
So, let's just keep that in mind in case we notice some issue later.
)

virxkane: > but I'm not really familiar with the lastFit handling.

I don’t know, this is mostly your code. But after analyzing it, it seemed to me that you simply did not leave unprocessed characters, so that's why. Now this is no longer the case and I did it by analogy with a simple FreeType render.

crengine/src/chmfmt.cpp
```diff
@@ -11,6 +11,10 @@
 #include "../include/crlog.h"
 #include <chm_lib.h>
 
+#include "../include/fb2def.h"
+#define XS_IMPLEMENT_SCHEME 1
+#include "../include/fb2def.h"
```
poire-z, line 16: This looks odd, as we have this comment:
https://github.com/buggins/coolreader/blob/bec2becc7ea8e34daf96a32e19f3da9810751eb9/crengine/include/fb2def.h#L4-L10

But I guess it's the easier way - and also, it will be local to chmfmt.cpp/chmfmt.o and shouldn't cause any issue ?

(The alternative would be to pick the _elementNameTable & co from the _`doc` passed from lvdocview to CHMTOCReader (which has them) and add them to the temporary `doc` LVParseCHMHTMLStream creates - but this looks more complicated - and more work for a format we don't really care about :)

virxkane: Yes, we include in chmfmt.cpp next items:
```cpp
static elem_def_t fb2_elem_table [] = { ... };
static attr_def_t fb2_attr_table [] = { ... };
static ns_def_t fb2_ns_table [] = { ... };
```
Ok, for alternative we must add getters, setters for `lxmlDocBase::_elementNameTable`, `lxmlDocBase::_attrNameTable`, `lxmlDocBase::_nsNameTable` and pass it to temporary `doc` in the `LVParseCHMHTMLStream()` function. Well, if you think that it will be more correct - do it. Later, as usual, I sync the code.

poire-z: I as much as you don't really want to do it, so go ahead with this :)
(It will add 20KB to libcrengine.so, which is not much.)

(Had a quick try with the following, as there is:
```c
/// Copy constructor - copies ID tables contents
lxmlDocBase::lxmlDocBase( lxmlDocBase & doc )
/// creates empty document which is ready to be copy target of doc partial contents
ldomDocument::ldomDocument( ldomDocument & doc )
: lxmlDocBase(doc)
```
but then, it would to do too much and crash when wanting to init styles on this new document... So, let's forget that option :)
```diff
@@ -861,3 +857,3 @@ public:

-ldomDocument * LVParseCHMHTMLStream( LVStreamRef stream, lString32 defEncodingName )
+ldomDocument * LVParseCHMHTMLStream( LVStreamRef stream, lString32 defEncodingName, ldomDocument * parent_doc )
 {
@@ -903,8 +899,6 @@ ldomDocument * LVParseCHMHTMLStream( LVStreamRef stream, lString32 defEncodingNa
     ldomDocument * doc;
-    doc = new ldomDocument();
+    doc = new ldomDocument(*parent_doc);
     doc->setDocFlags( 0 );
-    doc->setNodeTypes(fb2_elem_table);
-    doc->setAttributeTypes(fb2_attr_table);
-    doc->setNameSpaceTypes(fb2_ns_table);

+    printf("new doc blockquote id=%d\n", doc->getElementNameIndex(U"blockquote"));
     ldomDocumentWriterFilter writerFilter(doc, false, HTML_AUTOCLOSE_TABLE);
@@ -1106,3 +1100,3 @@ public:
             }
-            ldomDocument * doc = LVParseCHMHTMLStream( tocStream, defEncodingName );
+            ldomDocument * doc = LVParseCHMHTMLStream( tocStream, defEncodingName, _doc );
             if ( !doc ) {
```

poire-z: Just to get back these 20KB, I might go with:
```diff
--- a/crengine/include/lvtinydom.h
+++ b/crengine/include/lvtinydom.h
@@ -1300,2 +1300,4 @@ public:
     void setNameSpaceTypes( const ns_def_t * ns_scheme );
+    // set node/attribute/namespace types by copying them from an other document
+    void setAllTypesFrom( lxmlDocBase * d );

diff --git a/crengine/src/chmfmt.cpp b/crengine/src/chmfmt.cpp
index 1f4f23f6..4b7a80ad 100644
--- a/crengine/src/chmfmt.cpp
+++ b/crengine/src/chmfmt.cpp
@@ -7,6 +7,2 @@

-#include "../include/fb2def.h"
-#define XS_IMPLEMENT_SCHEME 1
-#include "../include/fb2def.h"
-
 #define DUMP_CHM_DOC 0
@@ -861,3 +857,3 @@ public:

-ldomDocument * LVParseCHMHTMLStream( LVStreamRef stream, lString32 defEncodingName )
+ldomDocument * LVParseCHMHTMLStream( LVStreamRef stream, lString32 defEncodingName, ldomDocument * parent_doc )
 {
@@ -905,5 +901,3 @@ ldomDocument * LVParseCHMHTMLStream( LVStreamRef stream, lString32 defEncodingNa
     doc->setDocFlags( 0 );
-    doc->setNodeTypes(fb2_elem_table);
-    doc->setAttributeTypes(fb2_attr_table);
-    doc->setNameSpaceTypes(fb2_ns_table);
+    doc->setAllTypesFrom(parent_doc);

@@ -1106,3 +1100,3 @@ public:
             }
-            ldomDocument * doc = LVParseCHMHTMLStream( tocStream, defEncodingName );
+            ldomDocument * doc = LVParseCHMHTMLStream( tocStream, defEncodingName, _doc );
             if ( !doc ) {
diff --git a/crengine/src/lvtinydom.cpp b/crengine/src/lvtinydom.cpp
index e5e69b78..9e0fa14e 100644
--- a/crengine/src/lvtinydom.cpp
+++ b/crengine/src/lvtinydom.cpp
@@ -5119,2 +5119,23 @@ void lxmlDocBase::setNameSpaceTypes( const ns_def_t * ns_scheme )

+// set node/attribute/namespace types by copying them from an other document
+void lxmlDocBase::setAllTypesFrom( lxmlDocBase * d )
+{
+    // Simpler to just serialize and deserialize them all
+    SerialBuf buf(0, true);
+    d->_elementNameTable.serialize( buf );
+    buf << d->_nextUnknownElementId;
+    d->_attrNameTable.serialize( buf );
+    buf << d->_nextUnknownAttrId;
+    d->_nsNameTable.serialize( buf );
+    buf << d->_nextUnknownNsId;
+
+    buf.setPos(0);
+    _elementNameTable.deserialize( buf );
+    buf >> _nextUnknownElementId;
+    _attrNameTable.deserialize( buf );
+    buf >> _nextUnknownAttrId;
+    _nsNameTable.deserialize( buf );
+    buf >> _nextUnknownNsId;
+}
+
 void lxmlDocBase::dumpUnknownEntities( const char * fname )
```
if that's ok with you.
I checked that `printf("chm blockquote id=%d\n", doc->getElementNameIndex(U"blockquote"));` gives 26 (as with your code) instead of 512, so I guess it should fix your original auto close tags issue too.

poire-z: For info because it's a bit related, I've added this to some other cleanup work of mine (soon to be PR'ed):
```diff
--- a/crengine/src/lvdocview.cpp
+++ b/crengine/src/lvdocview.cpp
@@ -4651,6 +4651,17 @@ void LVDocView::createEmptyDocument() {
     m_doc->setNodeTypes(fb2_elem_table);
     m_doc->setAttributeTypes(fb2_attr_table);
     m_doc->setNameSpaceTypes(fb2_ns_table);
+    // Note:
+    // All book formats parsing start with this, and get these fb2def.h
+    // elements defined and included in the document ids maps. This allows
+    // all our specific XHTML DOM element handling in ldomDocumentWriter
+    // using "==el_body", "==el_head" to match on books.
+    // But secondary XML files (.opf, .ncx...) get parsed with
+    // LVParseXMLStream(), which will not inject these known ids maps
+    // and will start a DOM with empty maps: met elements and attributes
+    // will get IDs starting from 512 (even if named <body> or <head>),
+    // and won't trigger these specific elements handling where we
+    // use these < 512 IDs (el_body, el_head...), which is good!
 }
```

virxkane: It seems to work. Who makes the commit?  Me and then you copy or you and then I copy?

poire-z: As you wish.
I'll be making a PR later today with it among others, I'll just tag this one "(Upstream)" if you make a PR before me.
But if you wait for my PR to pick it - you'll have 20 other commits to pick to keep in sync, so it depends on when you want to do that painful koreader-sync work :/ There may also be some risky commits, so you might want to wait a bit before syncing.

PR #312, opened 2021-09-07
ZIP64 fixes & other...

virxkane:
ZIP64 fixes:
* Potential memory leak fixed in crengine/src/lvstream/lvziparc.cpp.
Introduced in 01181eb40532ee191a1f8cf4b1f430ddee7876d9 (PR #310).
Thanks to @poire-z : https://github.com/koreader/crengine/pull/454#issuecomment-913805804
* Allows to open/extract inner files beyond 4G limit.
* If a zip64 signature is found but 'large file support' is disabled, do not attempt to scan the zip as a corrupted file (if the OS is capable of handling large files).

Other:
* Fixed wxWidgets build failure: compilation error due to API change. Introduced in c47cab44799475a48230312b8e97cd7a71edddb0.

### DIFF
--- a/cr3wx/src/view.cpp
+++ b/cr3wx/src/view.cpp
@@ -307,33 +307,73 @@ void cr3view::OnTimer(wxTimerEvent& event)
     }
 }
 
+static bool getBatteryState(int& state, int& chargingConn, int& level)
+{
+#ifdef _WIN32
+    // update battery state
+    SYSTEM_POWER_STATUS bstatus;
+    BOOL pow = GetSystemPowerStatus(&bstatus);
+    if (pow) {
+        state = CR_BATTERY_STATE_DISCHARGING;
+        if (bstatus.BatteryFlag & 128)
+            state = CR_BATTERY_STATE_NO_BATTERY;  // no system battery
+        else if (bstatus.BatteryFlag & 8)
+            state = CR_BATTERY_STATE_CHARGING;    // charging
+        chargingConn = CR_BATTERY_CHARGER_NO;
+        if (bstatus.ACLineStatus==1)
+            chargingConn = CR_BATTERY_CHARGER_AC; // AC power charging connected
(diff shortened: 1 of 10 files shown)


poire-z:
(Put back as draft because you have doubts about the fix ? Or just because you plan on adding more (related? unrelated?) commits to this PR? Just wondering if I can merge mine or if I should wait :)

virxkane:
> (Put back as draft because you have doubts about the fix ? Or just because you plan on adding more

Plan adding some commits (or one) to fully (or not full but more closer) support of zip64 feature.

poire-z:
Just asking because your last changes are too big for me to dig into :)

> If zip64 signature found but 'large file support' is disabled don't try to scan zip as broken file.

Can we have a zip64 signature with < 4GB files and normal EPUBs/FB2.ZIP which are usually smaller - or is it assured you won't get that signature until you reach 4GB and the need for it ?
And so, do "scan zip as broken" will still always work with smaller files (if we don't #define 'large file support') ?

Also, is your first zip64 commit in the other PR + your fix in the first commit here still just fine, and I can pick them safely ? Or did you fix some of the stuff they did with the commits you just added ?
Just asking if I can go on with them in my PR, or better keep them out till I get a chance to pick them all and do again some testing :)

virxkane:
> Can we have a zip64 signature with < 4GB files and normal EPUBs/FB2.ZIP which are usually smaller

Usually not.

> or is it assured you won't get that signature until you reach 4GB and the need for it ?

Yes. Also if zip file is large than 4G, 'End of central directory record' have special version and it's also signature of the zip64.

> And so, do "scan zip as broken" will still always work with smaller files (if we don't #define 'large file support') ?

I think, yes.

> Also, is your first zip64 commit in the other PR + your fix in the first commit here still just fine, and I can pick them safely ? Or did you fix some of the stuff they did with the commits you just added ?

If you pick only this two commits, then you can't read/extract inner files beyond 4G limit in such archives.

PR #327, opened 2022-02-24
android: add intents for some reader actions (e.g.: pageUp/pageDown/ttsPlay)

teleshoes:
so, i use the intents in this PR for a bunch of things. i use half-press of my camera button in SFOS to turn pages in coolreader in waydroid.

i think it might be nice to expose (all/many/some) reader commands via intents or a service.

### DIFF
--- a/android/src/org/coolreader/CoolReader.java
+++ b/android/src/org/coolreader/CoolReader.java
@@ -741,7 +741,17 @@ public class CoolReader extends BaseActivity {
 			if (uri != null) {
 				fileToOpen = filePathFromUri(uri);
 			}
+		} else if ("org.coolreader.PAGE_UP".equals(intent.getAction())) {
+			mReaderView.onCommand(ReaderCommand.DCMD_PAGEUP, 0, null);
+			return true;
+		} else if ("org.coolreader.PAGE_DOWN".equals(intent.getAction())) {
+			mReaderView.onCommand(ReaderCommand.DCMD_PAGEDOWN, 0, null);
+			return true;
+		} else if ("org.coolreader.TTS_PLAY".equals(intent.getAction())) {
+			mReaderView.onCommand(ReaderCommand.DCMD_TTS_PLAY, 0, null);
+			return true;
 		}
+
 		if (fileToOpen == null && intent.getExtras() != null) {
 			log.d("extras=" + intent.getExtras());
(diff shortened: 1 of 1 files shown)


plotn:
@teleshoes , can you more detailed explain which way you use that intents?

teleshoes:
my phone is a Sony Xperia XZ2 Compact, running SailfishOS
(specifically, this community port: https://github.com/sailfishos-sony-tama/main)

the keyevent listener is a sailfish-specific utility i wrote that basically wraps `cat /dev/input/by-path/*gpio_keys*` and parses the output.
https://github.com/teleshoes/sf-button-monitor

teleshoes:
i suggest: `org.coolreader.command.<ENUM_NAME>` or `.readercommand` to avoid namespace collisions with any explicitly added intents

out of curiousity, what does knownreader do?
'rich UI andf number of unique features' is there a feature list? also, 5000 side-merged commits? you should really consider rebasing :D

plotn:
If you want to discuss KR's features - it is better to do in 4pda theme or in KR's github discussions tab

android/src/org/coolreader/CoolReader.java
```diff
@@ -741,7 +741,17 @@ private boolean processIntent(Intent intent) {
...
 			}
+		} else if ("org.coolreader.PAGE_UP".equals(intent.getAction())) {
+			mReaderView.onCommand(ReaderCommand.DCMD_PAGEUP, 0, null);
+			return true;
+		} else if ("org.coolreader.PAGE_DOWN".equals(intent.getAction())) {
+			mReaderView.onCommand(ReaderCommand.DCMD_PAGEDOWN, 0, null);
+			return true;
+		} else if ("org.coolreader.TTS_PLAY".equals(intent.getAction())) {
+			mReaderView.onCommand(ReaderCommand.DCMD_TTS_PLAY, 0, null);
+			return true;
```
plotn, line 752: I did it via:

for (ReaderAction ra: ReaderAction.AVAILABLE_ACTIONS) {
			String acname = "org.knownreder." + StrUtils.getNonEmptyStr(ra.id, true);
			if (acname.equals(intentAction)) {
				mReaderView.onCommand(ra.cmd, ra.param, null);
				return true;
			}
		}

I think this is more convient solution

teleshoes:
closing in favor of generic impl in #329 
thanks plotn!

PR #337, opened 2022-06-06
Thirdparty repo fixes

virxkane:
* thirdparty management: use shasum instead of sha512sum.
* thirdparty management: when downloading, instruct curl not to save the output on failure.
* thirdparty libjpeg: for any reason remote file jpegsrc.v9d.tar.gz is changed but it's contents is NOT. So it is safely to update SHA512 sum for this file.
* thirdpart zlib: bump to 1.2.12.

### DIFF
--- a/thirdparty-deploy.sh
+++ b/thirdparty-deploy.sh
@@ -83,7 +83,7 @@ deploy_package()
 		then
 			rm -f "${SRCFILE}"
 		fi
-		curl -L -O ${URL} || die "Failed to fetch sources!"
+		curl -f -L -O ${URL} || die "Failed to fetch sources!"
 		if [ ! -f "${SRCFILE}" ]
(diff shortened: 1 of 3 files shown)


gladykov:
On my system (Endevour OS) had to symlink sha512sum to shasum, as there is no such bin, despite Perl  5.36.0-1 is installed on my machine

virxkane:
> On my system (Endevour OS) had to symlink sha512sum to shasum, as there is no such bin, despite Perl 5.36.0-1 is installed on my machine

And the other distribution doesn't have sha512sum.
Try this patch: [deploy-tool-shasum-find.patch.zip](https://github.com/buggins/coolreader/files/10151100/deploy-tool-shasum-find.patch.zip)
Make a PR yourself if it helps.

PR #339, opened 2022-06-23
Trying to fix license collision

virxkane:
Related to #338.

* Added script `crengine/Tools/getauthors/getauthors.sh` to fetch authors list. Of course this is only parts of the authors. Will be used only for guidance when compiling the list of authors.
* Added AUTHORS file.
* With the permission of the authors of the project, update the README to specify the license clearly "GPLv2 or later".
* With the permission of the authors of the project, the wording of the license has been changed to a clear "GPLv2 or later" in Qt Desktop GUI.
* Update source file headers: use the standard GPLv2+ header.
* Try to reconstruct the authors for each source file and specify them in the header.
* Revert adding flavors 'fdroid' & 'primary'.
* eink-onyx-reflections android library has been removed since its license status is unknown. So we lost support for ONYX devices.
* Replace hyphenation dictionaries, use only GPLv2+ compatible dictionaries, LPPL is not allowed.
  * At the same time, the format of hyphenation dictionaries has been updated: added properties 'title', 'lang', 'lefthyphenmin', 'righthyphenmin' to eliminate hardcoding this data in sources.
* Write contributing guidelines: indicate that contributors must add their own authorship to the AUTHORS file and to the header of the modified source file.
* Remove orphans files/code.


buggins:
@virxkane thank you!
Merged.

PR #342, opened 2022-08-31
page header in scroll mode + tts

### LINKED ISSUE #326: top+bottom margins ignored in scroll view
top+bottom margin works in page-view-mode only. the values have no effect in scroll-view-mode.
left+right margins work in both page-view and scroll-view.

teleshoes:
also fixes #326

### DIFF
--- a/android/src/org/coolreader/crengine/ReaderView.java
+++ b/android/src/org/coolreader/crengine/ReaderView.java
@@ -2497,10 +2497,6 @@ public class ReaderView implements android.view.SurfaceHolder.Callback, Settings
 			case DCMD_PAGEUP:
 				if (isBookLoaded()) {
 					boolean animationEnabled = pageFlipAnimationMode != PAGE_ANIMATION_NONE;
-					if (!mIsPageMode) {
-						//PAGEUP animation is broken in ViewMode=SCROLL, so skip it
-						animationEnabled = false;
-					}
 					if (animationEnabled && param == 1 && !DeviceInfo.EINK_SCREEN) {
 						animatePageFlip(-1, onFinishHandler);
 					} else {
@@ -3835,14 +3831,10 @@ public class ReaderView implements android.view.SurfaceHolder.Callback, Settings
 							BackgroundThread.instance().executeGUI(onFinishHandler);
 					}
 				} else {
-					//new ScrollViewAnimation(startY, maxY);
-					int fromY = dir > 0 ? h * 7 / 8 : 0;
-					int toY = dir > 0 ? 0 : h * 7 / 8;
-					new ScrollViewAnimation(fromY, h);
+					new ScrollViewAnimation(dir > 0 ? h*7/8 : 0-(h*7/8));
 					if (currentAnimation != null) {
 						nextHiliteId++;
 						hiliteRect = null;
-						currentAnimation.update(w / 2, toY);
 						currentAnimation.move(speed, true);
 						currentAnimation.stop(-1, -1);
 						if (onFinishHandler != null)
@@ -4137,7 +4129,8 @@ public class ReaderView implements android.view.SurfaceHolder.Callback, Settings
 //						sx = 0;
 				new PageViewAnimation(sx, maxX, dir);
 			} else {
-				new ScrollViewAnimation(startY, maxY);
+				int dir = newX < startX || newY < startY ? -1 : 1;
+				new ScrollViewAnimation(dir * currPos.pageHeight * 7 / 8);
 			}
 			if (currentAnimation != null) {
 				nextHiliteId++;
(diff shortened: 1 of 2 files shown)


plotn:
feature [crengine: add pageHeader and top/bottom margins in SCROLL mode (and TTS)]
is very requestable from the community, but it has bugs (maybe only in KR, maybe not) - when sliding scroll mode by finger (track mode) header is painted unexpectedly (in the middle).
And when I set position of titlebar = bottom, then it fails completly. I cannot merge such behaviour onto KR, too buggy. Could you make better? 
all other commits I skipped, cause I did the similar work, and I think we did nearly the same.

teleshoes:
@plotn
oh, whoops, i didnt know about either of those features. the titlebar=footer issue is straight-forward, but possibly a lot of work, since it doubles the use cases (especially annoying for page animation).

how do you enable "sliding scroll mode by finger" in coolreader? does it work by taking the pre-rendered page image and animating it forward/back?

teleshoes:
also, for TTS, the TTS control overlay may cover it. i could add a few px spacer, maybe. ill think about it when i get some time

plotn:
> sliding scroll mode by finger

Nothing special - put finger down, a little bit wait, then move finger up and down. The page scrolls accordingly finger moving

teleshoes:
is that a knownreader-specific feature? i have never seen smooth-scrolling finger sliding in coolreader.

plotn:
[cr3.ini.txt](https://github.com/buggins/coolreader/files/11295949/cr3.ini.txt)

no, this is common feature. Try my cr3.ini (dont forget to rename it)

teleshoes:
oh, you just mean gesture page flipping. that doesnt scroll, it just does multiple page turns at different speeds. if you do 1-page, it turns the page slowly. (you can also do gesture left/right instead of up/down)

the 'issue' you are referring to ALSO happens when you turn pages by tapping. the animation is just faster.

in fact, the same thing happens in PAGE mode, with or without this pull request.
in page mode, the animation is left-to-right, and in scroll mode, the animation is top-to-bottom.

so, in page mode, you will see the title bar slide to the left/right whenever you turn a page with animations enabled. it does NOT stay still, it is not treated specially. its just part of the page.

in scroll mode, it just slides to the top/bottom. either way, its considered part of the page, and the animation transition includes it. the fact that it slides up/down makes it look surprising, but the behavior is actually consistent.

this is, truly, not a bug.

teleshoes:
also: titlebar=footer is definitely a bug, i didnt know the titlebar could be on the bottom. it might be tricky to fix it.

but its not a REGRESSION. you ALREADY cannot have titlebar=footer in scroll mode, this pull request doesnt change that.

you should be willing to merge this, as-is. (i will still add the titlebar=footer fix when i get time, but possibly not very soon)

plotn:
Lets discuss, please.
Case 1. Scroll mode. I tap once at a bottom half of the screen, page picture scrolls with animation from bottom to top, then it stops, hearer is drawn. Everything is fine!

Case 2. Scroll mode. I tap and hold and begin to move finger from bottom to top and page goes visually at the distance of finger move, but! the header is drawn nearly in the center and it looks weird.

I can imagine the mechanics of this, but it should not be as it is, I think, isnt it ? Maybe there should be some variations of "get page image from the engine to android part" - when sliding the page, then without header, when release, then finally draw the header.

teleshoes:
i believe you are misunderstanding a few things:

1) case1 and case2 are not ACTUALLY different cases; they just have different animation speeds.
there is *NOT* any feature in coolreader where you actually *scroll* with your finger. gesture-flip just lets you *turn pages* with your finger, at varying speeds, following how fast you drag your finger.
try changing gesture from '1 page' to '10 pages', and you will see it flipping pages faster than normal, instead of slower than normal.

i tried this 'finger scroll' with your cr3.ini, and it was just doing normal `Gesture Page Flipping`.

2) case1 is not quite correct. the header is not drawn when it stops. the header travels up from the bottom of the page, through the middle, to the top, just like in case2.


if you are sure that i am wrong about either of these, we can discuss further, but i believe that this is a misunderstanding, not a bug. (i DO agree that it looks weird, though. i say that it is consistent, correct, and ugly.)

plotn:
I completely agree and understand the mechanics, but it behaviour and look in some cases ugly - yes, it is.

And I thinking about something like this: the drag event consists fro three 'steps' - tap (and hold), drag, release.

what if:

- when we tap we (re)draw page without header;
- while we drag - header is off;
- on release - redraw with header.

?

Maybe some flag to "prepare page image" or something else?

teleshoes:
1) when i first implemented this, i tried doing that. implementing it is easy, but it causes the words to instantly slide around, ruining the point of animation. i considered replacing the titlebar with a margin, so that it would be empty space that slides through the middle of the screen, but that isnt much less ugly.

you would have to: render current page, disable title bar, offset the current position by ~30px, render next page, adjust offset, start animation, enable title bar, select next page.

or we could change that way that animation is done and move the rectangles to clip the area of the titlebar. this is quite tricky, though.

2) most importantly, i dont think it belongs in this pull request. the animations are **correct** as-is, and changing their function might improve the look of the ui, but it breaks consistency with page mode, where we do not hide the titlebar during animation. this pull request tries to have feature-parity with page mode.

plotn:
Okay, okay.
Looks... as it looks. I respect. Sure, community will accept. My own compassion - I will make it disablabe in KR, ok?

Btw, tested new audiobook fixes. It is great! Dont you mind if I make an artilce on the popular russian IT resource - habr?

teleshoes:
yea, i intend to add a few more options to coolreader as well, when i get time, including this one (and a gui option for autoscroll speed). if i improve the animation, ill add options for that as well.

sure! please attribute, and mention that its GPLv2+ in the article. just a heads up, im going to be updating the main perl script soon, to replace `pandoc` with the new `cr3qt` api.

plotn:
I'll call option PROP_STATUS_HEADER_IN_SCROLL_MODE, agreed?

plotn:
By the way - there is one feature request from one user - when tap on header line (page number) - open dialog "go to page".
Can you help me to determine weather tap was at header line?

teleshoes:
the related options are called 'titlebar' or sometimes just `title`
e.g.: checkbox labeled `options_page_show_titlebar_title` sets property `window.status.title`, which is called `PROP_SHOW_TITLE` in the code.
(one more ridiculous thing in this codebase)

i recommend PROP_SHOW_TITLEBAR_IN_SCROLL_MODE=`window.status.title.scrollmode` or something similar

teleshoes:
for this unrelated feature, i think you should simply add additional tap zones.

here is what you could do:
1) add 3 more tap zones (9+10+11), with y=TITLEBAR_HEIGHT, and x=.33/.66/1.0. steal this area from tap zones 0+1+2, or shave it off the top and divide equally the remaining area. (id recommend the former)
2) add an option, disabled by default, `[] Include separate titlebar tap zones`
3) when disabled, in the edit-tap-zone actions, either re-calculate the tap zones, or just copy the action from zones 0+1+2 to the new titlebar zones at the top 9+10+11
4) when enabled, in the ui, show 12 tap zones

for the record, i hate this feature. tapping the top right corner should not need to be a precision action. that is why i recommend 3 tap zones, so that the enduser can have just the middle one show go-to-page.
have you suggested that the user just do long-tap in the top-middle (zone1) or even short tap in the middle? 18 tap actions should be enough, and its not like showing go-to when you click the titlebar is an intuitive feature that should be present by default.

teleshoes:
@buggins @plotn
TL/DR there is no bug here that i am aware of, and this pull request can be merged as-is, without regression, to the best of my knowledge.

this PR fixes a bug (animation in scroll mode) and implements an unexpectedly missing feature. with this PR, titleBar/pageHeader/pageFooter works consistently and correctly in scroll mode and page mode, at the top or bottom, and with or without animation.

before this PR, titleBar was ignored in scroll mode, and animations in scroll mode were buggy.

notes:
1) you cannot see the scroll-mode page footer in TTS mode, because the android navigation bar covers it. (this is already the case in page mode when you show the android navigation bar)
2) the animation includes the title bar, like it does in page mode. if this is visually unappealing, an alternative can be implemented, in both page and scroll mode, but it is unrelated to this pull request.

PR #357, opened 2023-04-30
android: add a simple state log file

teleshoes:
writes a single line at startup, shutdown, tts start, and tts stop. it currently writes to the coolreader app dir, but could write to external storage for interfacing with other apps.

i use this to control coolreader TTS audio using the volume up/down buttons while my phone is locked. (if coolreader is running and in TTS mode, play/pause coolreader. otherwise, play/pause my media player).

i have previously used a combination of querying the android media layer, pgrep to see if tts is running, and logcat, which worked a little less reliably and a lot less efficiently.


teleshoes:
@buggins this one is also safe to merge, i've been using it continuously for about 3 years, on multiple devices for several people

PR #365, opened 2025-04-19
Set CMake policy version lower bound to 3.5

penguinpee:
Also specify upper bound as 4.0.

With CMake 4.0 policies before 3.5 will be dropped. CMake 4.0 also
requires the upper bound to be specified.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,5 +1,5 @@
 PROJECT(cr3)
-cmake_minimum_required(VERSION 3.0)
+cmake_minimum_required(VERSION 3.5..4.0)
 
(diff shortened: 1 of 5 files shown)


penguinpee:
Fedora is planning to upgrade CMake to 4.0.

I tested the Linux build with CMake 4.0 as well as CMake 3.30.8 and didn't notice any regression. I did _not_ test the Android build nor any of the third party libraries (we do not use them). I just assumed they would be compatible.

More info:

https://cmake.org/cmake/help/latest/release/4.0.html
https://cmake.org/cmake/help/latest/manual/cmake-policies.7.html#unsupported-policies

Jappie3:
thanks for figuring this out!
I've fixed the NixOS package for coolreader by vendoring this PR as a patch: https://github.com/NixOS/nixpkgs/pull/466281

PR #375, opened 2026-07-20
Issue 372 upgrade thirdparty libraries

buggins:
```
Thirdparty library versions for local builds upgraded:
	freetype      2.14.3
	fribidi       1.0.16
	harfbuzz      14.2.1
	jpeg          10
	libpng        1.6.58
	libunibreak   7.0
	zlib          1.3.2
	zstd          1.5.7
	utf8proc      2.11.3
```

PR #376, opened 2026-07-20
thirdparty libs - fix android sources + definitions

teleshoes:


teleshoes:
for me, i need this pr to build android at all after #375
each cmake sources change fixes a missing symbol or obsolete flag
here are some of the errors fixed:
- harfbuzz: Cannot find source file: thirdparty/harfbuzz-14.2.1/src/hb-ot-shape-complex-arabic.cc
- libpng: PNG_ARM_NEON_CHECK_SUPPORTED must not be defined on ARM64
- freetype: error: undefined symbol: FT_Set_Named_Instance

teleshoes:
(i see you added some of these already, so i rebased and left just the ones i had and you didn't. i don't know if this set is still necessary)

Issue #378, opened 2026-07-22
CREngine fixes to match best practices for modern C++


buggins:
Add performance measurement tool to give visibility for impact from changes.
Cover library with tests.
Update implementation according to modern C++ standards.

virxkane:
Maybe replace `LVRef<>` with `std::shared_ptr<>` & other similar things?

buggins:
@virxkane -- yes, and many more.

PR #381, opened 2026-07-25
Crengine optimizations 2026 pt1

buggins:
It's going to be a big fix with rework of crengine core parts.
issue-#378


pkb:
It doesn't look very modern; it seems COW isn't used these days. Why not just use std::string and std::string_view? I'm not a C++ expert, though.

